PR #5, opened 2015-11-19
project reorganized so that all deps are managed on one place

ppitonak:
* all dependencies are managed on one place
* non-production dependencies will be pruned during "gulp generate" phase
* all non-production dependencies need to be defined in devDependencies section of package.json
* tests removed from distribution


kenfinnigan:
merged

PR #6, opened 2015-11-20
Add squirrel event handling

jrichter1:
So that once the app is packaged using squirrel, it doesn't wait ~10 seconds on every launch and then restart.


kenfinnigan:
merged

PR #10, opened 2015-11-26
Update unit tests and minor tweaks

jrichter1:


browser/services/data.js
```diff
@@ -15,6 +13,7 @@ class InstallerDataService {
     this.downloading = false;
     this.installing = false;
 
+    let env = require('remote').require('../main/env');
```
kenfinnigan: Is this move purely to make the declaration of 'env' closer to its use or another reason?

jrichter1: Yes, there is another reason - when I tried to import the class into the tests I got the following:
```Cannot find module 'remote'```
This seems to have solved it.

kenfinnigan:
Is this PR needed anymore?

jrichter1:
The updated unit tests should still be relevant, but I need to change them again. I can either do it as a part of this PR, or I can close this and create another one when I'm finished.

kenfinnigan:
Whatever is easier for you is fine

PR #14, opened 2015-12-02
tests for account controller

ppitonak:
@kenfinnigan please review, test for JDK is broken so it needs to be disabled/removed, run tests with "npm test"


kenfinnigan:
merged

PR #16, opened 2015-12-04
Working tests for jdk install

jrichter1:
first batch of 'model' tests


browser/model/jdk-install.js
```diff
@@ -34,7 +34,7 @@ class JdkInstall extends InstallableItem {
       }
     };
 
-    request(options)
+    request.get(options)
```
kenfinnigan: Is there a particular reason this was changed?

jrichter1: The ```request.get``` can be stubbed or spied, which makes it a bit nicer to test. It does the same thing, doesn't it? Because some other classes use get and some use just the request. It might looks better if they are unified.

PR #18, opened 2015-12-11
Attempt to refactor downloading and move common code into a separate class

jrichter1:


kenfinnigan:
Was this tested at runtime?

From looking at the code, I'm not even sure it would work as the cdk model has multiple downloads, and the handler would just overwrite the stream for each one. I could be wrong, but that's what it looks like.

I like the idea though, as I was thinking of pulling out a common request handler at some point

jrichter1:
I've just run the installer, and it works! :)
It looks like the downloads proceed in a sequential manner, so the stream gets overwritten after the previous one was closed.

kenfinnigan:
At the moment all downloads within cdk are parallel. They're not super huge right now but I'd prefer not to switch them being done sequentially

jrichter1:
Ok now I'm a bit puzzled - I've added some logging to the download method, and it starts all 10 downloads right at the beginning. So they still do run in parallel. But how the streams work is something I have no clue about.

kenfinnigan:
I know why now. After each write stream call there is a download() or downloadAuth() call.

It's those download() calls which actually create the request instance, after which point the request doesn't access the writestream on the Download class anymore, so setting it to a new one doesn't cause an overwrite.

That's my presumption anyway

kenfinnigan:
Though error handling and end of stream could cause it to close the wrong one, unless there's a closure around the request call that retains the value of writeStream from when download() was called.

jrichter1:
If that was the case, wouldn't end of stream cause it to always close the stream that was declared as the last one? Example: one stream starts -> another stream starts -> the first stream ends - at this point the field should point to the second stream, so the end method should close the second stream even if it hasn't ended yet. But this doesn't happen. So either magic happens and the code works, or I got extremely lucky.

kenfinnigan:
Right, so I suspect there's a closure surrounding the request() call that retains the variables on the download handler in their state at the time of the call, which would mean writeStream would remain the value at the time request() was called.

Would need to use debugger to verify that though.

Can you verify that? If so, I think we can merge this

jrichter1:
I could not find what I was looking for, so I made the download methods retain the stream themselves.

PR #27, opened 2016-01-15
Allow to enable DevTools in installer window

dgolovin:
Use

``` shell
export PDKI_DEBUG=
```

to enable DevTools or

``` shell
unset PDKI_DEBUG

```
to disable it.


dgolovin:
released to master

PR #29, opened 2016-01-20
Promisify virtualbox and test it

jrichter1:


dgolovin:
Released to master with removed code duplication.

PR #33, opened 2016-03-02
JBDS-3657 Installer should set updated default username and password in .cdk

alexeykazakov:


fbricon:
+1 for the change. Not convinced this causes the travis build failure though

alexeykazakov:
This build reports are useless right now. It fails all the time ;)

jrichter1:
The travis builds don't run properly because of github rate limit + jspm + travis only allowing to use a github token for branches in this repo. Builds for PRs from forks just won't work as of now, so don't worry about these. However, that does not mean you shouldn't check it locally :)

fbricon:
Applied in master

PR #34, opened 2016-03-08
JBDS-3615 Run virtualbox with elevated privileges

jrichter1:


jrichter1:
Node-windows breaks the tests, closing until a better solution is found.

jrichter1:
Found a way to get it working

PR #35, opened 2016-03-08
JBDS-3615 Run virtualbox with elevated privileges (round 2)

jrichter1:


joshuawilson:
Please update to the latest master branch.

dgolovin:
Works for me, except there is no log file anymore. @jrichter1 could you bring log back?

dgolovin:
There two problems:
- I am getting "Machine entered the wrong state" error constantly with this PR;
- Using node-windows elevate method exits right away after running runas command and this makes virtualbox installation marked as finished in installer whereas actual installer is still running as separate process.

jrichter1:
The first issue seems to be environment related (and I'm not getting it still) - I found a whole slew of reasons this why could happen (vt-x disabled, some VM service already running, conflicting PATH entries, etc.)

The second problem (that I've overlooked in the first place, good catch) is very much in node-windows design. The only way to find out when the process has finished is via the callback passed in, but it uses a 'bootstrap' command. And the callback is called when the bootstrap terminates. Right now I have no idea how to mend this - might have to drop node-windows.

ppitonak:

dgolovin:
I did, see my comment above, with node-windows privilege elevation it is not acceptable.

dgolovin:
I also tried https://www.npmjs.com/package/runas for this, but it doesn't work for windows 10.

dgolovin:
we can try to do that with poweshell script:
``` poweshell
Start-Process msiexec.exe "/qn /norestart /i c:\installer.msi INSTALLDIR=
c:\location /log c:\location\log.txt" -Verb runas -Wait
```
this script triggers normal privilege lifting dialog for Windows Installer.

dgolovin:
This PR https://github.com/redhat-developer-tooling/developer-platform-install/pull/38:
- lifts privileges with -Verb runas and -Wait let us hold "Installing" until installer is finished.

jrichter1:
Closing as there is another solution.

PR #39, opened 2016-03-17
JBDS-3647 Investigate why we are including both cygwin and mingw

dgolovin:
returning cygwin back for these reasons:
- it is the only ssh/rsync implementation that can be used to
  configure sshd service and then use version 1.0.0 of vagrant-sshfs
  to   mount the users home directory into the VM (see JBDS-3671 for
  details) ;
- it works with latest vagrant release 1.8.1, mingw based ssh/rsync
  does not. This is the going tombe a problem in case someone has it
  already installed and decided to skip vagrant installation and
  use preinstalled one.


ppitonak:

PR #40, opened 2016-03-18
JBDS-3682 Online installer reports missing package.js file

dgolovin:
None pretty solution v.2 to fix the tests and app at the same time.


ppitonak:

PR #41, opened 2016-03-18
JBDS-3685 simple example of zipping output installer folder into a si…

nickboldt:
…ngle zip; next up, trying to do this with 7z self-extracting


ppitonak:

dgolovin:
That is good start :) Now wee need to download 7zip extension, CDK, JDBS, Java put it in the same folder dist\win\DeveloperPlatformInstaller-win32-x64 with generated electron app and zip it as self Executable archive.

nickboldt:
Note: msi installer runs and then does nothing. But the .exe works great! 

Do we need the .msi? Do we need the .zip? I can remove them if not required here.

installers.json
```diff
@@ -0,0 +1,20 @@
+{
+  "installerURLs": {
+    "cdk" : {
+      "0" : "http://cdk-builds.usersys.redhat.com/builds/03-Mar-2016/cdk.zip",
```
alexeykazakov: Some human-readable names instead of numbers would look better.

nickboldt: Though the indices are arbitrary & meaningless, I've added some labels you might enjoy instead.

dgolovin:
Nick, thanks for this PR :+1:  I merged it, with some refactoring:
1. Renamed installers.json to requirements.json and fixed loading at runtime;
2. Changed structure of jason file to simple plain map;
3. Removed electron installer related tasks, because 'generate' task does the same;
4. Removed some other changes that is not related to bundled installer.

What is left to do is figure out how to download and install 7zip on jenkins, then download and install 7zip-extensions with sfx modules, then run it to zip results of generate and prefetch.

nickboldt:
" Removed electron installer related tasks, because 'generate' task does the same;"

Actually, my electronwinstaller task doesn't do exactly the same as your "generate" task: mine uses an animated gif while the installer loads, and contains more files within the .exe than yours.

Mine is also a single self-contained .exe; yours is a folder full of .dlls. 

Haven't tested if mine can be produced without the need for the generate step. If so, mine might be a cleaner solution as it produces fewer files in the output folder.

dgolovin:
You are talking about package task, which produces single .exe file, which in turn the same selfextracting archive by nature 🙂

PR #42, opened 2016-03-18
Added POC UI test script - for installer login failure

ldimaggi:


ppitonak:

ppitonak:
test this

ppitonak:
@jrichter1 please review

test/ui/login-test.js
```diff
@@ -0,0 +1,79 @@
...
+let PAGE_TITLE = 'Red Hat Developer Platform Installer';
+let BAD_USERNAME = "badusername1";
+let BAD_PASSWORD = "badpassword1";
+var loginButton;
+
+describe('Login', function() {
+
+  beforeEach(function() {
+    loginButton = protractor.until.elementLocated(By.className('btn-lg'))
+  });
```
jrichter1: This would be better suited as a beforeAll function - you only need to find the element once.

test/ui/login-test.js
```diff
@@ -0,0 +1,79 @@
...
+    webdriver.wait(loginButton, 10000)
+    .then(function(elm) {
+      expect(elm.isEnabled()).toEqual(false);
+    });
+  });
+
+describe('Bad username and password', function() {
+  beforeEach(function() {
+    var usernameField = element(by.id('username'));
+    var passwordField = element(by.id('password'));
```
jrichter1: Only one retrieval of the elements is needed - would be better to move the element discovery to a local beforeAll so it doesn't have to repeat for every test case.

dgolovin:
Pushed to master.

PR #43, opened 2016-03-20
JBDS-3690 Update to CDK Beta5

alexeykazakov:
~~Still need to check if we now can get rid of the Vagrantfile provided by the https://github.com/redhat-developer-tooling/openshift-vagrant~~
Update: Update to CDK Beta5 + use Vagrantfile provided by CDK 
See https://issues.jboss.org/browse/JBDS-3656

### DIFF
--- a/browser/main.js
+++ b/browser/main.js
@@ -91,10 +91,9 @@ let mainModule =
                 CDKInstall.key(),
                 new CDKInstall(installerDataSvc,
                                 $timeout,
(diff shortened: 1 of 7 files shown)


jrichter1:
Please rebase on master and let the verification job run again.

alexeykazakov:
Updated the PR. CDK is updated to Beta5 + Vagrantfile provided by CDK is used.

PR #46, opened 2016-03-22
JBDS-3656 The installer should use Vagrantfile provided by CDK.

alexeykazakov:
Use cdkv2 instead of cdk_v2.


dgolovin:
released to master

PR #50, opened 2016-03-23
use released...

nickboldt:
use released /openshift-origin-client-tools-v1.1.4-3941102-windows.zip instead of lastSuccessfulBuild

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -105,7 +105,7 @@ gulp.task('ui-test', function() {
 });
 
 gulp.task('default', function() {
-  return runSequence('generate', 'create-zip', 'electronwinstaller');
+  return runSequence('generate', 'create-zip');
 });
 
 // download all the installer dependencies so we can package them up into the .exe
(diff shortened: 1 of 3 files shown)


gulpfile.js
```diff
@@ -121,7 +120,7 @@ gulp.task('prefetch', function() {
             var fname = body.match(/openshift-origin-client-tools-v\w(\.\w)(\.\w){1,3}-\w{1,3}-\w{8}-\w{7}-windows\.zip/)[0];
             console.log('DOWNLOADING -> ' + currentUrl.concat(fname));
             request(currentUrl.concat(fname))
-              .pipe(fs.createWriteStream(path.join(prefetchFolder, currentKey)));
+              .pipe(fs.createWriteStream(path.join(prefetchFolder, key)));
```
dgolovin: This doesn't work with callbacks.

dgolovin:
picked one commit with cdk-test changes and released to master.

ppitonak:
@nickboldt for the future, could you please include only related commits in PRs? e.g. this PR is about updating OpenShift client tools but there is also commit "remove ref to electronwinstaller as task no longer exists"

nickboldt:
@ppitonak I can do more PRs if you want but figured that a handful of commits in the same PR would be OK. Sometimes it's hard refrain from fixing typos when I see them, relegating them to a second PR

PR #51, opened 2016-03-23
JBDS-3704 The installer gets stuck on Download page and never goes to Getting Started

alexeykazakov:


dgolovin:
released to master.

PR #52, opened 2016-03-23
JBDS-3653 two tasks to show how we might...

nickboldt:
JBDS-3653 two tasks to show how we might convert long paths with spaces to their 8.3 short paths - use 'gulp hd' to check your home dir, or 'gulp bc' to convert a long folder path


dgolovin:
this issue was solved by showing warning with explanation how to fix it.

PR #58, opened 2016-03-24
JBDS-3669 version the .exe

nickboldt:
JBDS-3669 version the .exe using either the package.json version


nickboldt:
Rebased. See also https://github.com/redhat-developer-tooling/developer-platform-install/pull/74 for more on parameterizing the output .exe filename. These two changes will need to be carefully merged.

ppitonak:
So with this change, we have this two cases:

1. I can set version in package.json and it will appear inside installer but `npm run package` will produce exe without version or
2. I can run it with `gulp package --version 0.0.2-something` but it will only be part of exe name but won't be visible inside installer

Why don't you just use the version from package.json inside gulpfile to rename exe?

nickboldt:
Simpler PR drops the idea of commandline override. Better?

dgolovin:
Haven't tested it, but it should work IMO.

ppitonak:
Merged.

PR #63, opened 2016-03-31
Avoid jspm 0.16.32 which causes build failures

jrichter1:
Jspm 0.16.32 causes build and run errors right now - workaround to revert to 0.16.31 until a proper solution is found.


ppitonak:
Merged manually.

PR #65, opened 2016-03-31
wrap node/gulp build with a maven pom so we...

nickboldt:
wrap node/gulp build with a maven pom so we can run it in Jenkins more easily

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1,0 +1,165 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  Copyright 2016 Red Hat, Inc., and individual contributors
+-->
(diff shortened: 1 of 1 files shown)


dgolovin:
I don't think we need it until we really going to build installer as part of something bigger that uses maven, otherwise it adds two extra dependencies: java and maven.

maxandersen:
why do you need a maven build to link jobs together in jenkins ? not following that since jenkins jobs can be linked together no matter the used build step inside it.

> I should note too that if you don't want to use the pom to build... Don't. You can still manually install npm & gulp then run that way.

lets not add extra layers of build stuff if it is not needed. Right now I don't see a need for it so please enlighten us why introducing maven is needed - jenkins itself don't need maven to do builds.

ppitonak:
It seems that adding Maven is not useful. Can we close this?

PR #77, opened 2016-04-04
JBDS-3706 - Update the Patternfly lib to 3.3.2.

joshuawilson:


joshuawilson:
Looks like the build failed. I know that for this to build locally, I needed to remove the node-modules, jspm modules, and the config.js file. They all get auto-generated. 

I don't want to break the build so I am not going to merge this to master until someone can remove those.

ppitonak:
@joshuawilson, Jenkins job does remove _node_modules_, _browser/config.js_ and _browser/jspm_packages_. IMHO the problem is the following error during _npm install_

    > bootstrap-treeview@1.2.0 install /mnt/hudson_workspace/workspace/developer-platform-installer-pr/node_modules/bootstrap-treeview
    > bower install

    sh: bower: command not found

We don't use bower to install dependencies, any idea why it fails?

jrichter1:
This is an issue introduced in Patternfly 3.1.0. It should be resolved in 3.3.2 so I suggest you try upgrading to 3.3.2 and see what happens.

ppitonak:
Upgrading to 3.3.2 doesn't help, I can see this issue open https://github.com/patternfly/patternfly/issues/236

joshuawilson:
@ppitonak I did have to install bower and then I went to Patternfly in the node-modules and did a npm install too.

joshuawilson:
I removed node-modules, jspm-modules, and the config.js. I then ran npm install and I got the bower install error. That is because the first time I had installed it locally. I removed everything again and installed bower globally. Then npm install worked and everything displays fine. I also updated the PR to 3.3.2.

ppitonak:
To summarize:

- we would like to upgrade to latest version of Patternfly (3.3.2 right now)
- there is a bug during _npm install_ that require Bower being installed globally when you (transitively) depend on bootstra-treeview (patternfly/patternfly#236, jonmiles/bootstrap-treeview#196)

Solutions:

1. We can either run _npm install -g bower_ before _npm install_ or
2. we can stay on Patternfly 2.x until the bug is fixed

I would prefer not to introduce unnecessary global dependencies but  if we decide for this solution, we need to make sure that the dependencies are updated properly as soon as possible.

joshuawilson:
I just added bower to the internal install and it appears to solve the problem. There is now no need to add a global dependency on bower or install it locally.

joshuawilson:
Rebased with latest from master and merged.

PR #80, opened 2016-04-04
JBDS-3761 ensure that existing VBox install...

nickboldt:
JBDS-3761 ensure that existing VBox install is detected and like with Vagrant, is selected for reuse by default


dgolovin:
Good catch :) thanks!

dgolovin:
Merged with rebase and minor change.
Thanks again :)

PR #74, opened 2016-04-02
parameterized .exe filename ("" or "-bundle")

nickboldt:
parameterized .exe filename ("" or "-bundle"); readme addendum moved to #82

PR #81, opened 2016-04-04
better default gulp targets

nickboldt:
add package-simple, package-bundle, and dist; update README and
package.json; remove create-zip (not needed); by default do 'run', not build.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -48,13 +48,6 @@ gulp.task('clean', function() {
   return del(['dist'], { force: true });
 });
 
-// currently not used
-gulp.task('create-zip', () => {
-  return gulp.src(buildFolderRoot + buildFolderRoot + '/**/*')
-      .pipe(zip(buildFolderRoot + '.zip'))
-      .pipe(gulp.dest(buildFolderRoot));
-});
-
 // Create default callback for exec
 function createExecCallback(cb, quiet) {
   return function(err,stdout,stderr) {
@@ -66,7 +59,7 @@ function createExecCallback(cb, quiet) {
   }
 }
 
-gulp.task('generate', ['clean', 'transpile:app'], function(cb) {
+gulp.task('generate', ['transpile:app'], function(cb) {
   var electronVersion = pjson.devDependencies['electron-prebuilt'];
   var cmd = path.join('node_modules', '.bin') + path.sep + 'electron-packager . ' + artifactName + ' --platform=' + artifactPlatform + ' --arch=' + artifactArch;
   cmd += ' --version=' + electronVersion + ' --out=./' + buildFolderRoot + ' --overwrite --asar=true';
@@ -74,6 +67,9 @@ gulp.task('generate', ['clean', 'transpile:app'], function(cb) {
 
   exec(cmd,createExecCallback(cb, true));
 });
+
+// default task
+gulp.task('default', ['run']); 
 
 gulp.task('run', ['transpile:app'], function(cb) {
   exec(path.join('node_modules', '.bin') + path.sep + 'electron .',createExecCallback(cb));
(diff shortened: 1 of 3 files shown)


nickboldt:
with the above changes, you can produce the stub installer and the bundle installer like this:

    gulp
    mv dist/win/*.exe dist/win/Setup-stub.exe
   gulp package-bundle
   mv dist/win/*.exe dist/win/Setup-bundle.exe

But if someone merges https://github.com/redhat-developer-tooling/developer-platform-install/pull/74 then the two move steps are no longer needed.

ppitonak:
This PR looks good, the only thing I don't like is "stub" in name of new Gulp task. @nickboldt @dgolovin @jrichter1 any suggestions for better name?

nickboldt:
How about package-simple and package-bundle? I could also add a package-all that would build both in a single default operation (clean, generate + package both installers, then cleanup 7zip leftovers).

Note that doing both in one task requires that someone push https://github.com/redhat-developer-tooling/developer-platform-install/pull/74 so that the output files from each package task don't overwrite each other.

ppitonak:
I'm testing #74 right now, download of all deps takes too long :(

nickboldt:
After this change, running just 'gulp' gets this:

$➔ ll dist/win/
total 1.6G
-rwxr-xr-x 1 nboldt 197609  37M Apr  6 11:24 DeveloperPlatformInstaller-win32-x64.exe
drwxr-xr-x 1 nboldt 197609    0 Apr  6 11:24 DeveloperPlatformInstaller-win32-x64
-rwxr-xr-x 1 nboldt 197609 1.6G Apr  6 11:50 DeveloperPlatformInstaller-win32-x64-bundle.exe

dgolovin:
I got both installers 1.6G after building.

ppitonak:
1. I suggest that we rename _packages_ task to _dist_
2. you should update npm scripts in package.json to contain start (default task), package-simple, package-bundle and dist
3. you should update Readme
3. I suggest that we change default task to start _run_ task as this is what people probably run during development

nickboldt:
Should the default task be "start" or "run" ? You appear to be asking for both.

gulpfile.js
```diff
@@ -126,9 +119,19 @@ gulp.task('package', function (cb) {
   });
 });
 
-// Create bundled installer
+// Create stub installer that will then download all the requirements
```
dgolovin: all package* tasks should have clean as first task in sequence.

gulpfile.js
```diff
@@ -170,3 +169,7 @@ gulp.task('prefetch', function(cb) {
   }
   artifactType = "-bundle";
 });
+
+gulp.task('default', function() {
```
dgolovin: Should be

``` JavaScript
gulp.task('default', ['run'], function() {});
```

gulpfile.js
nickboldt: I disaagree. Why do you need to clean again after creating the first installer?

gulpfile.js
nickboldt: But transpile (which is what the run task does) is meaningless w/o first doing a generate, right?

gulpfile.js
ppitonak: I think that you are not talking about the same thing :) 

1. clean should definitely run before task _package-simple_ and _package-bundle_
2. in task _packages_, it should run as first task, doesn't need to run after first installer is created

gulpfile.js
ppitonak: This is what I asked for in my comment

    gulp.task('default', ['run']);  // nothing more is necessary

Transpile creates main/index.js, it has nothing to do with generate

nickboldt:
Refactored again. Is this what you both want? Wasn't sure about adding "start" to package.json and if we still need to include a "package":"gulp package" line.

ppitonak:
Package.json already contains _start_ script and it calls `gulp run`, I'm fine with having it this way. For the future, please verify that Markdown files render correctly ;)

Merged. Thanks for your PR.

PR #84, opened 2016-04-05
Fix jspm install so the builds may succeed

jrichter1:
This immediate patch is to fix the build failures on npm install. Note that it is supposed to be a temporary solution to an in-depth problem which would require a more complex solution that we should defer to after the initial release.


ppitonak:
Merged to master, please fix test.

PR #86, opened 2016-04-05
instead of -stub and -bundle, use '' and...

nickboldt:
instead of -stub and -bundle, use '' and -bundle to differentiate the two installers (@dgolovin's suggestion)


nickboldt:
oops. not sure why this happened

PR #89, opened 2016-04-06
Update the progress directive tests to include the new fields

joshuawilson:


joshuawilson:
I know these tests fail. But I am unable to run them as I get the following error.

Error: EMFILE: too many open files

I made the start of the changes to the tests. So at least the basic structure should be right.

dgolovin:
On my side everything works, so windows distribution should be fine. I fixed minor issues and released in master.

ppitonak:
The error with too many open files is indirectly caused by Patternfly 3

PR #97, opened 2016-04-06
JBDS-3778 2 methods: getSHA256(filename) & createSHA256File(filename)

nickboldt:
JBDS-3778 2 methods: getSHA256(filename) & createSHA256File(filename)

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -2,6 +2,7 @@
 
 var gulp = require('gulp'),
     fs = require('fs-extra'),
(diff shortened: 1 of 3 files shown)


nickboldt:
This was working great w/ 100% sync methods until I tried it on something bigger than the cygwin.exe file. So, I've added a bunch of async methods but line 233 isn't working yet. 

Need some help/suggestions how to make the async call properly do the isExistingSHA256Current() check and return true into the prefetch task on line 250.

PR #106, opened 2016-04-08
JBDS-3710 add metadata and into the electron...

nickboldt:
JBDS-3710 add metadata and into the electron installer exe file

PR #110, opened 2016-04-08
JBDS-3710 fix icon + metadata in the...

nickboldt:
JBDS-3710 fix icon + metadata in the electron exe, then use ResourceHacker to do the same for the 7zip installer exe


nickboldt:
Supercedes #106 - contains the same fix (for the electron .exe) but also fixes the DevPlatformInstaller-*.exe file too.

PR #121, opened 2016-04-11
Check requirements URLs before building

jrichter1:
To avoid certain situations that have arisen throughout the development, this PR aims to check the requirements before building so the urls:
 - respond with 200
 - point to a file with roughly the size we expect
 - contain some text specific to the tool that should be downloaded (i.e. vagrant url contains 'vagrant' - we've been there before :) )

### DIFF
--- a/test/check-requirements.js
+++ b/test/check-requirements.js
@@ -1,0 +1,84 @@
+'use strict';
+
+let reqs = require('../requirements.json');
+let request = require("request");
+
+let minSizes = new Object();
+let data = new Object();
+let fileNames = new Object();
+let count = 0;
+
+function checkRequirements() {
+  for (var attribute in reqs) {
+    data[attribute] = reqs[attribute].url;
+    count++;
+  }
+
+  //to check if the url looks like it points to what it is supposed to
+  fileNames['cdk.zip'] = 'cdk';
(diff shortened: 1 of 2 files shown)


test/check-requirements.js
```diff
@@ -0,0 +1,84 @@
+'use strict';
+
+let reqs = require('../requirements.json');
+let request = require("request");
+
+let minSizes = new Object();
```
ppitonak: why don't you use simply `let minSizes = {};`?

test/check-requirements.js
```diff
@@ -0,0 +1,84 @@
...
+  if (!data[key].includes(fileNames[key])) {
+    throw new Error(key + ' - the url does not contain ' + fileNames[key] + ': ' + data[key]);
+  }
+}
+
+function checkUrl(key) {
+  let req = request.get(data[key])
+  .on('response', function(response) {
+    let size = response.headers['content-length'];
+    if (response.statusCode !== 200) {
```
ppitonak: What happens if the URL will redirect to a new place?

jrichter1: Both virtualbox and vagrant use redirects, something like this happens:
virtualbox.exe - url: http://developers.redhat.com/redirect/to/virtualbox-5.0.8.download
virtualbox.exe - size: 117095112B

ppitonak:
Merged

PR #123, opened 2016-04-12
Fix vagrant setup commands not all running in sequence

jrichter1:
Vagrant setup reports complete before 'setx VAGRANT_DETECTED_OS "cygwin"' is called. This PR should fix that.


ppitonak:
Merged

PR #124, opened 2016-04-12
current 7zip cleanup leaves behind 7zS.exe;...

nickboldt:
current 7zip cleanup leaves behind 7zS.exe; this gets everything and doesn't block (force = false)


ppitonak:
verifying right now

ppitonak:
Merged

PR #132, opened 2016-04-13
JBDS-3638 newer cdk.zip; .box file appears...

nickboldt:
JBDS-3638 newer cdk.zip; .box file appears to be the same

### DIFF
--- a/requirements.json
+++ b/requirements.json
@@ -1,12 +1,12 @@
 {
   "cdk.zip": {
     "bundle": "yes",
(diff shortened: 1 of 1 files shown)


alexeykazakov:
Changing requirements.json is not enough. You also should update vagrant-service-manager-1.0.0.gem to vagrant-service-manager-1.0.1.gem - https://github.com/redhat-developer-tooling/developer-platform-install/blob/master/browser/model/cdk.js#L232
See an example here: https://github.com/redhat-developer-tooling/developer-platform-install/pull/108/files

alexeykazakov:
Replaced by https://github.com/redhat-developer-tooling/developer-platform-install/pull/137

PR #143, opened 2016-04-14
move prefetched requirements to a new folder...

nickboldt:
move prefetched requirements to a new folder so they can be reused between builds

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ test/jspm-config.js
 .idea
 xunit.xml
 browser-tests.xml
+prefetch-dependencies/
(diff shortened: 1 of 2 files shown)


ppitonak:
- [x] `npm run package-simple` works fine
- [ ] `npm run package-bundle` with no prefetch deps downloads all deps and generates *36MB installer*
- [ ] `echo 'aaa' >> prefetch-dependencies/cdk.zip && npm run package-bundle` correctly detected that file is corrupt and started downloding it again, Gulp task _prefetch_ failed because of "task completion callback called too many times"
- [ ] `npm run dist` - didn't run this

nickboldt:
Yeah, I know about "task completion callback called too many times" but not sure how to fix. Will squash if someone can suggest how to fix the callback problem.

ppitonak:
BTW, it would be nicer to implement prefetch task using Promise.all (https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Promise/all)

dgolovin:
@ppitonak @nickboldt +1 for using Promise.all

nickboldt:
replaced by #158

PR #158, opened 2016-04-25
JBDS-3849 use promises to fetch requirements

nickboldt:
JBDS-3849 use promise to fetch requirements into cache folder and copy from cache to dist

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -145,9 +145,9 @@ gulp.task('create-7zip-archive', function(cb) {
   // only include prefetch folder when zipping if the folder exists and we're doing a bundle build
   if (fs.existsSync(path.resolve(prefetchFolder)) && installerExe.indexOf("-bundle") > 0)
   {
-    packCmd = packCmd + ' ' + path.resolve(prefetchFolder) + path.sep + '*';
+    packCmd = zaExe + ' a -x!*.sha256 ' + bundled7z + ' ' + zaElectronPackage + path.sep + '*' + ' ' + path.resolve(prefetchFolder) + path.sep + '*';
   }
-  //console.log('[DEBUG]' + packCmd);
+  console.log('[INFO] ' + packCmd);
   exec(packCmd, createExecCallback(cb, true));
 });
 
@@ -227,7 +227,7 @@ gulp.task('package-simple', function(cb) {
 
   // Create bundled installer
 gulp.task('package-bundle', function(cb) {
-  runSequence(['check-requirements', 'clean'], ['prefetch','create-dist-win-dir'], ['generate',
+  runSequence(['check-requirements', 'clean'], 'prefetch','create-dist-win-dir', ['generate',
     'prepare-tools'], 'package', 'cleanup', cb);
 });
 
@@ -277,47 +277,33 @@ function resolveInstallerExePath(artifactType) {
 
 // prefetch all the installer dependencies so we can package them up into the .exe
 gulp.task('prefetch', function(cb) {
-  let promises = [];
   if (!fs.existsSync(prefetchFolder)) { mkdirp(prefetchFolder); }
+  let counter=0;
   for (var key in reqs) {
     if (reqs.hasOwnProperty(key)) {
       let currentUrl = reqs[key].url;
       let currentKey = key;
       let currentFile = path.join(prefetchFolder, key);
       let alreadyDownloaded = false;
-      promises.push(new Promise((resolve,reject)=>{
       // if file is already downloaded, check its sha against the stored one
-        isExistingSHA256Current(currentFile, reqs[currentKey].sha256sum, (alreadyDownloaded)=> {
-          //console.log('For ' + currentFile + ": " + alreadyDownloaded + ", " + reqs[currentKey].bundle);
-          if(!alreadyDownloaded) {
-            // download only what can be included in offline installer
-            if (reqs[currentKey].bundle === 'yes') {
-              promises.push(new Promise(function(resolve, reject) {
-                console.log('[INFO] Download ' + currentUrl + ' to ' + currentFile);
-                request(currentUrl)
-                  .pipe(fs.createWriteStream(currentFile)).on('finish',function(err, res) {
-                    if (err) { 
-                      reject(err);
-                    } else {
-                      //console.log('[DEBUG] Downloaded ' + currentFile);
-                      createSHA256File(currentFile);
-                      resolve(res);
-                    }
-                  });
-              }));
+      isExistingSHA256Current(currentFile, reqs[currentKey].sha256sum, (alreadyDownloaded)=> {
+        //console.log('For ' + currentFile + ": " + alreadyDownloaded + ", " + reqs[currentKey].bundle);
+        if(!alreadyDownloaded && reqs[currentKey].bundle === 'yes') { // download only what can be included in offline installer
+          console.log('[INFO] Download [' + counter + '] ' + currentUrl + ' to ' + currentFile);
+          counter++;
+          request(currentUrl).pipe(fs.createWriteStream(currentFile)).on('finish',function() {
+            createSHA256File(currentFile);
+            counter--;
+            if(counter===0) {
+              installerExe = resolveInstallerExePath('-bundle');
+              cb();
             }
-          }
-        });
-      }));
+          });
+        }
+      });
     }
   }
-  Promise.all(promises).then((result) => {
-    installerExe = resolveInstallerExePath('-bundle');
-    cb();
-  }).catch((error)=>{
-    cb(error);
-  });
-
+  installerExe = resolveInstallerExePath('-bundle');
 });
 
 //check if URLs in requirements.json return 200 and generally point to their appropriate tools
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -28,7 +28,7 @@ var artifactName = 'jboss-devstudio-platform',
 var buildFolderRoot = 'dist/win/';
 var buildFileNamePrefix = artifactName + '-' + artifactPlatform + '-' + artifactArch;
 var buildFolder = buildFolderRoot + buildFileNamePrefix;
-var prefetchFolder = buildFolderRoot + buildFileNamePrefix; // or just use downloads/ folder to that a clean doesn't wipe out the downloads
+var prefetchFolder = 'requirements'; // use folder outside buildFolder so that a 'clean' task won't wipe out the cache
 let buildFolderPath = path.resolve(buildFolderRoot);
 
 let zaRoot = path.resolve(buildFolderRoot);
@@ -52,7 +52,7 @@ gulp.task('transpile:app', function() {
     .pipe(gulp.dest('./main'));
 });
 
-// clean dist/ AND downloads/ folder
+// clean dist/ AND prefetch-dependencies/ folder
 gulp.task('clean-all', ['clean'], function() {
   return del([prefetchFolder], { force: true });
 });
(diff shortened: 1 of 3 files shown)


gulpfile.js
```diff
@@ -286,22 +308,37 @@ gulp.task('prefetch', function(cb) {
...
-          console.log('DOWNLOADED <- ' + currentFile);
+          counter++;
+          promises[counter] = copyFromPrefetchToDist(currentFile, destFile);
+          counter--;
         }
       });
+
     }
   }
+  promise.all(promises);
```
dgolovin: Nick it should be something like this snipped below and coounter would not be neded anymore 
``` JavaScript
promise.all(promises).then((result)=> {
  cb();
}).catch((error)=>{
  cb(error);
});
```

requirements.json
```diff
@@ -12,7 +12,7 @@
   "oc.zip": {
```
dgolovin: Please send separate PR for this change.

nickboldt:
SHA fix for requirements moved to #159 at Denis' request.

nickboldt:
So, this PR works if you run `gulp prefetch && gulp dist`

But if you start from a clean checkout (no prefetch folder) or do `gulp clean-all && gulp dist` the sequence looks like this:

http://pastebin.com/7jg3merG

gulpfile.js
```diff
@@ -269,14 +270,34 @@ function resolveInstallerExePath(artifactType) {
...
+    });
+  }
+  return aPromise;
+}
+
+// prefetch all the installer dependencies so we can package them up into the .exe
 gulp.task('prefetch', function(cb) {
-  let counter=0;
+  let promises = [];
+  if (!fs.existsSync(prefetchFolder)) { mkdirp(prefetchFolder, cb); }
```
dgolovin: This is the reason why, you call back here and the in promise.all().then().catch();

nickboldt:
With this last suggestion, build works.

Here's a console log (expires in 1 week):

http://pastebin.com/WpAjCp2U

dgolovin:
Fix changes a lot and not easy to follow, instead is should be simple promise wrap around download we already have. Here is an example https://github.com/redhat-developer-tooling/developer-platform-install/compare/master...dgolovin:JBDS-3849?expand=1.
I didn't test it really and it could look much better after refactoring, but I think that is Pavol's idea in rough implementation. It just get rid of ugly counter and introduce Promise.all() call.

gulpfile.js
```diff
@@ -17,6 +17,7 @@ var gulp = require('gulp'),
     minimatch = require('minimatch'),
     copy = require('gulp-copy'),
     concat = require('gulp-concat'),
+    promise = require('promise'),
```
ppitonak: Promises are natively supported by Node.js 4 and 5 so you don't need this dependency.

gulpfile.js
```diff
@@ -269,14 +270,34 @@ function resolveInstallerExePath(artifactType) {
   return path.join(zaRoot, artifactName + '-' + pjson.version + artifactType + '-installer.exe');
 }
 
-// download all the installer dependencies so we can package them up into the .exe
+function copyFromPrefetchToDist(currentFile, destFile) {
```
ppitonak: I see no reason why we need to copy all downloaded bits. We create 7z archive like this `7za.exe a installer.7z installer-dir/*`, can't we just add second path to this command? I would be both faster and would save 1GB on hard disk

`7za.exe a installer.7z installer-dir/* prefetch-dependencies/*`

gulpfile.js
```diff
@@ -79,16 +80,16 @@ gulp.task('create-dist-win-dir', function(cb) {
...
+  var cmd = path.join('node_modules', '.bin') + path.sep + 'electron-packager . ' + artifactName + ' --platform=' + artifactPlatform + ' --arch=' + artifactArch;
   cmd += ' --version=' + electronVersion + ' --out="' + buildFolderPath + '" --overwrite --asar=true';
   cmd += ' --version-string.CompanyName="Red Hat, Inc."';
   cmd += ' --version-string.ProductName="' + pjson.productName + '"';
   cmd += ' --version-string.OriginalFilename="' + artifactName + '-' + pjson.version + '-installer.exe"';
   cmd += ' --version-string.FileDescription="' + pjson.description + ' v' + pjson.version + '"';
   cmd += ' --app-copyright="Copyright 2016 Red Hat, Inc."';
   cmd += ' --app-version="' + pjson.version + '"' + ' --build-version="' + pjson.version + '"';
-  cmd += ' --prune --ignore=test';
+  cmd += ' --prune --ignore="test|prefetch-dependencies"';
```
dgolovin: This seems is not working for me, so I got all the prefetch-dependencies inside first light installer..

gulpfile.js
```diff
@@ -285,23 +306,33 @@ gulp.task('prefetch', function(cb) {
...
+
     }
   }
+  promise.all(promises).then((result)=> {
+    cb();
+  }).catch((error)=>{
+    cb(error);
+  });
+
   installerExe = resolveInstallerExePath('-bundle');
```
ppitonak: This should go inside then statement.

gulpfile.js
```diff
@@ -285,23 +306,33 @@ gulp.task('prefetch', function(cb) {
...
           }
         } else {
-          console.log('DOWNLOADED <- ' + currentFile);
+          promises.push(copyFromPrefetchToDist(currentFile, destFile));
         }
       });
+
     }
   }
+  promise.all(promises).then((result)=> {
```
ppitonak: You need to put `return` statement in front of promise.all so that Gulp handles sequence of tasks correctly.

package.json
```diff
@@ -64,8 +64,10 @@
     "karma-junit-reporter": "0.4.1",
     "karma-ng-html2js-preprocessor": "0.2.1",
     "minimatch": "3.0.0",
+    "mkdirp": "^0.5.1",
     "mocha": "2.4.5",
     "mock-fs": "3.8.0",
+    "promise": "^7.1.1",
```
ppitonak: not necessary (see comment above)

package.json
```diff
@@ -64,8 +64,10 @@
     "karma-junit-reporter": "0.4.1",
     "karma-ng-html2js-preprocessor": "0.2.1",
     "minimatch": "3.0.0",
+    "mkdirp": "^0.5.1",
```
ppitonak: Not really necessary with Node.js 5.x but has to be listed explicitly with Node.js 4.x. I created a PR https://github.com/redhat-developer-tooling/developer-platform-install/pull/160 but we can add it with this PR, it doesn't matter.

ppitonak:
I added comments directly to Nick's code. I still need to try it

gulpfile.js
```diff
@@ -215,25 +216,25 @@ function createSHA256File(filename, cb) {
...
+  runSequence(['check-requirements', 'clean'], 'create-dist-win-dir', ['generate',
     'prepare-tools'], 'package', 'cleanup', cb);
-});
+});
 
   // Create bundled installer
 gulp.task('package-bundle', function(cb) {
-  runSequence(['check-requirements', 'clean'], 'create-dist-win-dir', ['generate',
+  runSequence(['check-requirements', 'clean'], 'create-dist-win-dir', ['generate',
     'prepare-tools'], 'prefetch', 'package', 'cleanup', cb);
```
ppitonak: we should be able to run _prefetch_ in parallel with _generate_ and _prepare-tools_, it will speed up the build a little bit

gulpfile.js
```diff
@@ -215,25 +216,25 @@ function createSHA256File(filename, cb) {
...
-});
+});
 
 // Create both installers
 gulp.task('dist', function(cb) {
-  runSequence(['check-requirements', 'clean'], 'create-dist-win-dir', ['generate',
+  runSequence(['check-requirements', 'clean'], 'create-dist-win-dir', ['generate',
     'prepare-tools'], 'package', 'prefetch', 'package', 'cleanup', cb);
```
ppitonak: we should be able to run _prefetch_ in parallel with _generate_ and _prepare-tools_, it will speed up the build a little bit

gulpfile.js
```diff
@@ -285,23 +306,33 @@ gulp.task('prefetch', function(cb) {
...
-              });
+            promises.push(new Promise(function(resolve, reject) {
+              console.log('[INFO] Download ' + currentUrl + ' to ' + currentFile);
+              request(currentUrl)
+                .pipe(fs.createWriteStream(currentFile)).on('finish',function(err, res) {
+                  if (err) { 
+                    reject(err);
+                  } else {
+                    // console.log('DOWNLOADED <- ' + currentFile);
+                    promises.push(copyFromPrefetchToDist(currentFile, destFile));
```
ppitonak: I think that this is causing trouble... you can't push copy promise to _promises_ array because it will be called in parallel with download task. You need to refactor it out and in the end do something like `return Promise.all(downloadFiles).then(copyFiles).then(() => cb()).catch(cb);`

gulpfile.js
ppitonak: Can you try adding parentheses? `--ignore="(test|prefetch-dependencies)"`

gulpfile.js
nickboldt: You mean before the cb() on line 331 ?

gulpfile.js
dgolovin: that is doable and should work good, because download is going to be done often for jbds bits only, probably

gulpfile.js
dgolovin: good point, it should work

gulpfile.js
ppitonak: yes

gulpfile.js
ppitonak: I would leave it as is for now and improve performance in another PR

package.json
nickboldt: I disagree. I get "ReferenceError: promise is not defined" if I don't declare it in gulpfile.js.

gulpfile.js
nickboldt: I disagree. I get "ReferenceError: promise is not defined" if I don't declare it in gulpfile.js.

ppitonak: you need to use Promise (with upper-case "P")

gulpfile.js
```diff
@@ -269,40 +275,49 @@ function resolveInstallerExePath(artifactType) {
...
-      });
+        });
+      }));
     }
   }
-  installerExe = resolveInstallerExePath('-bundle');
+  Promise.all(promises).then((result) => {
+    installerExe = resolveInstallerExePath('-bundle');
+    cb();
+  }).catch((error)=>{
```
dgolovin: should be just 
``` JavaScript
}).catch(cb);
```

package.json
dgolovin: promise => Promice should fix the issue.

gulpfile.js
```diff
@@ -269,40 +275,49 @@ function resolveInstallerExePath(artifactType) {
...
-        } else {
-          console.log('DOWNLOADED <- ' + currentFile);
-        }
-      });
+        });
+      }));
     }
   }
-  installerExe = resolveInstallerExePath('-bundle');
+  Promise.all(promises).then((result) => {
```
dgolovin: you should add return 
``` JavaScript
 return Promise.all(promises).then((result) => {
```

dgolovin:
https://github.com/redhat-developer-tooling/developer-platform-install/pull/161 reworked version of this PR

dgolovin:
See https://github.com/redhat-developer-tooling/developer-platform-install/pull/161 for details.

PR #159, opened 2016-04-25
update SHAs and use DNS-friendly host...

nickboldt:
update SHAs and use DNS-friendly host download-ipv4.eng.brq.redhat.com instead of download.devel.redhat.com for the jdk.zip

### DIFF
--- a/requirements.json
+++ b/requirements.json
@@ -12,7 +12,7 @@
   "oc.zip": {
     "bundle": "yes",
     "url": "https://github.com/openshift/origin/releases/download/v1.1.6/openshift-origin-client-tools-v1.1.6-ef1caba-windows.zip",
-    "sha256sum": "5d29a8725785dfc783dc3d0768931462f2cc1f47f09acf232f79b03a3ef4edb2"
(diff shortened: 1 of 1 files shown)


requirements.json
```diff
@@ -22,7 +22,7 @@
   "cygwin.exe": {
     "bundle": "yes",
     "url": "https://cygwin.com/setup-x86_64.exe",
-    "sha256sum": "08079a13888b74f6466def307a687e02cb26fc257ea2fa78d40f02e28330fd56"
+    "sha256sum": "58f9f42f5dbd52c5e3ecd24e537603ee8897ea15176b7acdc34afcef83e5c19a"
```
ppitonak: This checksum is not correct, the original one was fine.

ppitonak: Actually, there is different sha256 sum on Linux and Windows, we should investigate why

dgolovin: this is windows app and windows sha256 should be used, but it is interesting why. It uses Cryptographic hash function to calculate hash string and if message (.exe file) is the same, the message digest should also be the same.

ppitonak: The new sum seems to be correct... there was something wrong with my configuration

ppitonak:
It seems to be outdated now.

PR #165, opened 2016-05-05
JBDS-3878 pull installer version from...

nickboldt:
JBDS-3878 pull installer version from package.json

### DIFF
--- a/browser/model/jbds-autoinstall.js
+++ b/browser/model/jbds-autoinstall.js
@@ -12,7 +12,7 @@ class JbdsAutoInstallGenerator {
   }
 
   generate(jbdsInstallDir, jdkInstallDir) {
-    let jbdsVersion = "9.1";
+    let pjson = require('../../package.json');
     let temp =
       [
         '<?xml version="1.0" encoding="UTF-8" standalone="no"?>',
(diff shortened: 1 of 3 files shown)


nickboldt:
This PR is not quite perfect.... ran fine, built fine, even the tests ran OK. 

But on launch of the 1.6G installer, I get:

    system.js:4 Uncaught (in promise) Error: Error: ENOENT: no such file or directory, 
    open 'C:\Users\nboldt\AppData\Local\Temp\7zS82BB.tmp\package.json'(…)

So either need a different way to depend on package.json, or need to copy that file over into the .asar file somehow. @dgolovin @joshuawilson @jrichter1 @ppitonak any suggestion how to do that?

browser/model/jbds-autoinstall.js
```diff
@@ -12,7 +12,7 @@ class JbdsAutoInstallGenerator {
   }
 
   generate(jbdsInstallDir, jdkInstallDir) {
-    let jbdsVersion = "9.1";
+    let pjson = require('../../package.json');
```
dgolovin: DevStudio version should be loaded from requirements.json, not from package.json. It also wont work for tests and install time the same way, because root folder is different.

browser/model/jbds.js
```diff
@@ -7,6 +7,8 @@ let ipcRenderer = require('electron').ipcRenderer;
 let Glob = require("glob").Glob;
 let replace = require("replace-in-file");
 
+var pjson = JSON.parse(fs.readFileSync('./package.json'));
```
dgolovin: the same problem is here.

browser/model/jbds-autoinstall.js
nickboldt: OK, so I can switch to the other file and parse it out of the jbds.jar URL ... but that doesn't help with the other problem re: "no such file or directory" ... does it?

browser/model/jbds.js
nickboldt: Could I define a global variable in gulpfile.js and just use that in the browser/model/*.js files and tests?

dgolovin: I'll create Util.loadJsonFile(path) where path is relative path to json file from project root. It will check two possible locations for resources: development and runtime execution. Similar what is implemented for requirements.json loading here https://github.com/redhat-developer-tooling/developer-platform-install/blob/master/browser/main.js#L79.

browser/model/jbds-autoinstall.js
dgolovin: Could DevStudio product build just generate contents.json (similar to eclipse p2 repo) with some additional info
{
   "name": "Red Hat Developer Studio Nightly Build",
   "version": "10.0.0.Alpha2",
   "timestamp": "<buildQualifier>",
   "installer": "filename"
}

Installer could use base URL for contents.json and calculate url to download installer file and sha256 file for verification.

browser/model/jbds-autoinstall.js
```diff
@@ -38,15 +38,16 @@ class JbdsAutoInstallGenerator {
...
-        '<shortcut KdeSubstUID="false" categories="" commandLine="-jar &quot;' + path.join(jbdsInstallDir, 'Uninstaller', 'uninstaller.jar') + '&quot;" createForAll="false" description="Uninstall Red Hat JBoss Developer Studio ' + jbdsVersion + '" encoding="" group="true" icon="' + path.join(jbdsInstallDir, 'studio', 'jbds_uninstall.ico') + '" iconIndex="0" initialState="1" mimetype="" name="Uninstall Red Hat JBoss Developer Studio ' + jbdsVersion + '" target="' + path.join(jdkInstallDir, 'bin', 'javaw.exe') + '" terminal="" terminalOptions="" tryexec="" type="1" url="" usertype="0" workingDirectory=""/>',
+        '<programGroup name="Red Hat JBoss Developer Studio ' + pjson.version + '"/>',
```
dgolovin: this should be fixed at devstudio installer site, if <com.izforge.izpack.panels.ShortcutPanel id="shortcut"> is empty node, devstudio insaller should generated this as a default.

browser/model/jbds-autoinstall.js
nickboldt: You could also scrape the HTML page and get this information. 

https://devstudio.jboss.com/10.0/snapshots/builds/devstudio.product_master/latest/all/ -> jboss-devstudio-10.0.0.Alpha1-v20160506-2012-B5309-installer-standalone.jar

PR #167, opened 2016-05-05
JBDS-3605 JBDS-3878 fix branding & versioning

nickboldt:
JBDS-3605 fix branding (no longer JBoss)
JBDS-3878 fix versioning s/9.1/10.0/g

### DIFF
--- a/browser/model/jbds.js
+++ b/browser/model/jbds.js
@@ -18,8 +18,8 @@ import Util from './helpers/util';
 class JbdsInstall extends InstallableItem {
   constructor(installerDataSvc, downloadUrl, installFile, targetFolderName) {
     super('jbds',
-          'RED HAT JBOSS DEVELOPER STUDIO',
-          '9.1.0',
+          'RED HAT DEVELOPER STUDIO',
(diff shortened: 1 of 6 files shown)


browser/model/jbds.js
```diff
@@ -18,8 +18,8 @@ import Util from './helpers/util';
 class JbdsInstall extends InstallableItem {
   constructor(installerDataSvc, downloadUrl, installFile, targetFolderName) {
     super('jbds',
-          'RED HAT JBOSS DEVELOPER STUDIO',
-          '9.1.0',
+          'RED HAT DEVELOPER STUDIO',
```
ppitonak: This is the only component with name in upper case, is there any reason for that?

nickboldt: It was like that when I found it, I swear. :D

dgolovin: Should be "Red Hat Developer Studio" the same way as other components,

nickboldt:
Replaced by https://github.com/redhat-developer-tooling/developer-platform-install/pull/189

PR #171, opened 2016-05-09
Update controller unit tests

jrichter1:
Sadly, other controllers require electron built-in modules, so they cannot be tested on this level.


dgolovin:
released to master after rebase d3fe692

PR #173, opened 2016-05-09
Remove useless promise

jrichter1:
AFAIK it does nothing apart from making some unit tests fail on windows.

### DIFF
--- a/browser/model/virtualbox.js
+++ b/browser/model/virtualbox.js
@@ -64,9 +64,7 @@ class VirtualBoxInstall extends InstallableItem {
                 resolve(output);
             });
           } else {
-            return new Promise((resolve, reject) => {
-	             resolve(output);
-	            });
+	           resolve(output);
           }
         } else {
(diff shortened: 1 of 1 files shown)


browser/model/virtualbox.js
```diff
@@ -64,9 +64,7 @@ class VirtualBoxInstall extends InstallableItem {
                 resolve(output);
             });
           } else {
-            return new Promise((resolve, reject) => {
```
dgolovin: I remember this was exactly why detection for virtualbox didn't work for Nick.

jrichter1: I believe this is not the one Nick introduced - that one is still there (and it is commented).
See https://github.com/jrichter1/developer-platform-install/blob/promise-fix/browser/model/virtualbox.js#L82

dgolovin: I see the difference now, that one was inside that call. Current one is already inside Promise instance. You are right.

dgolovin:
+1 to merge

jrichter1:
Pushed to master.

PR #175, opened 2016-05-10
Fix test error under windows

dgolovin:
when running on windows there are extra method calls to mock


jrichter1:
Released to master. Thanks.

PR #181, opened 2016-05-13
JBDS-3666 Back off and retry if download fails

dgolovin:
Fix enables:
- downloader to save failed downloads
- instll step restart download from dialog

### DIFF
--- a/browser/model/cdk.js
+++ b/browser/model/cdk.js
@@ -59,17 +59,17 @@ class CDKInstall extends InstallableItem {
 
     let totalDownloads = 3;
 
-    let downloader = new Downloader(progress, success, failure, downloadSize, totalDownloads);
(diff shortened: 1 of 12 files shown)


jrichter1:
The ui test job does not look very stable yet. I can see no changes that would break it here, so let me launch it again and see what it does.

jrichter1:
test this

jrichter1:
Soo, the failure detection works for all the components, but the restart is only implemented for CDK, correct? That's going to throw errors any time something else fails to download and restart is clicked. Plus the VPN message is sort of CDK specific and hopefully only temporary.

dgolovin:
Yep, that was correct :) now it is fixed.

dgolovin:
Ouch, I tests were failing when request.defaults() used, fixed that and returned initial route to account in main.js.

PR #182, opened 2016-05-13
Increase login tests wait timeout

jrichter1:
It takes a bit more time on jenkins for the page to load, so give it more time.


dgolovin:
released to master after rebase

PR #185, opened 2016-05-17
JBDS-3889 Remove JBoss from Platform Installer full name

dgolovin:
Fix updates name to 'Red Hat Developer Studio PLatform Installer'


dgolovin:
npm install fails with error "Error: Cannot find module 'electron'" on CI host local build works fine

jrichter1:
Something interrupted the electron-prebuild download so the unit tests failed. As for the UI tests the path is currently hardcoded in the protractor config and I assume it needs adjusting. Could you do that as well?

dgolovin:
@jrichter1 thanks for pointing that out :) haven't run ui-tests.

dgolovin:
pr updated and all checks are good.

PR #187, opened 2016-05-17
Add missing requirements to url check

jrichter1:
so that the build might proceed


dgolovin:
released to master after re-base.

PR #193, opened 2016-06-01
JBDS-3787 Installer - Installation of each component should be optional

dgolovin:
Checkboxes added to select components to install
Autoselection for CDK and Devstudio Dependencies added


dgolovin:
replaced with https://github.com/redhat-developer-tooling/developer-platform-install/pull/206

PR #196, opened 2016-06-01
Switch to latest CDK 2.1 nightly build

dgolovin:


ppitonak:
@jrichter1 please review

jrichter1:
If it is CDK 2.1, can you also bump the version to 2.1 so it say it is installing CDK 2.1?
Otherwise no complaints from me.

dgolovin:
PR updated. CDK is now at v2.1. thanks @jrichter1!

PR #198, opened 2016-06-01
JBDS-3618 Progress does not reflect actual amount of work done

dgolovin:


dgolovin:
Not needed after https://github.com/redhat-developer-tooling/developer-platform-install/pull/297 is released.

PR #199, opened 2016-06-02
Get installables info from requirements.json

jrichter1:
Attempt to get all the names, versions and description from requirements.json instead of hardcoded strings in code.

### DIFF
--- a/browser/model/installable-item.js
+++ b/browser/model/installable-item.js
@@ -1,7 +1,8 @@
 'use strict';
 
-let path = require('path');
-let reqs = require(path.resolve('./requirements.json'));
+import Util from './helpers/util';
+
+let reqs = Util.resolveFile('.', 'requirements.json');
 
 class InstallableItem {
   constructor(keyName, installTime, downloadUrl, installFile, targetFolderName, installerDataSvc) {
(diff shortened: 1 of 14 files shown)


jrichter1:
Build is green, @dgolovin , @ppitonak , let me know what you think.

dgolovin:
@jrichter1 thanks, it looks good, but I would postpone releasing this right now, because of conflicts.

jrichter1:
I got it rebased for you :)

browser/model/installable-item.js
```diff
@@ -1,11 +1,23 @@
 'use strict';
 
+let path = require('path');
+let reqs = require(path.resolve('./requirements.json'));
```
dgolovin: this line only works for tests, it doesn't during installation.

dgolovin: Here is snippet that works https://github.com/redhat-developer-tooling/developer-platform-install/blob/master/browser/main.js#L79

jrichter1:
I just found out I've pushed a wrong commit here (whoops). Anyway, I got that fixed and rather than using the horrible piece of path resolving again I put it into a method.

dgolovin:
@jrichter1 could you rebase it one more time on top of all the changes?
Thanks!

jrichter1:
Looks like I've messed up somewhere, investigating.

jrichter1:
Rebased, mistakes mended.

PR #207, opened 2016-06-06
fix version of devstudio and use explicit...

nickboldt:
fix version of devstudio and use explicit URL w/ version


nickboldt:
Re-ran the build and it is green. https://dev-platform-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/developer-platform-installer-pr-build/47/

PR #211, opened 2016-06-06
JBDS-3926 cache 7zip, extras, and resource...

nickboldt:
JBDS-3926 cache 7zip, extras, and resource hacker in cache folder for reuse


jrichter1:
@nickboldt sorry for overlooking this for so long. If you could rebase it - you will also need to remove the resource hacker zip from the cleanup task - so we can merge it at last :)

jrichter1:
On second thought, it now looks easier to just replace this with a new PR and do it in a little bit different fashion. So I'll close this one.

PR #217, opened 2016-06-07
JBDS-3803 Add JavaVM version detection and validation

dgolovin:


dgolovin:
@jrichter1 could you try it out before I push it?  It is not part of BOM but a feature that is good to have.

jrichter1:
Will do.

jrichter1:
It seems unlike vagrant and virtualbox, java can be selected to install even if detected. Is that intended?

dgolovin:
That is correct, because OpenJDK is still delivered as a zip file, so we I unpack it  and then use in DevStudio installation without any conflicts with detected java.

jrichter1:
Good, I'm going to try some install combinations after my bundled build has finished downloading (which is probably going to take a while, sorry)

dgolovin:
Also notice if detected version is used there would be no progress on install page for openjdk, for vagrant and virtual box would be shown even if they are detected. It mostly because vagrant is always vagrant detected or installed. there are several reasons for different behavior here:
- vagrant detected or installed is still vagrant, where instead of OpenJDK we can use detected Oracle JDK;
- even if vagrant detected we still do environment configuration for it.

jrichter1:
Deselecting components causes an uncaught error when I click 'download and install' and everything basically falls apart. I'm using the build generated by the PR builder.

fail

dgolovin:
PR build might be the problem here. Could you take a look at https://issues.jboss.org/browse/JBDS-3787 using CI build form machydra?

jrichter1:
OK, the last commit is the cause. Builds up to 174 work, the newest doesn't.

dgolovin:
You right, that is an issue, with last change. I'll send PR with fix after I replicate it locally. 
Thanks Jan!

dgolovin:
https://github.com/redhat-developer-tooling/developer-platform-install/pull/219 to fix the issue. Problem is current UI is intermediate version on the way to original design with radio buttons. Current implementation with check boxes is using 'install' option value to indicate install is requested and 'detected' when install is skipped. 'detected' option is added to list of options only in case installer found already installed components, so code should check if 'detected' option is present before using it.

dgolovin:
https://github.com/redhat-developer-tooling/developer-platform-install/pull/219 is released to master

PR #219, opened 2016-06-08
Fix for missing version field

dgolovin:


dgolovin:
second build with the same PR is good.

PR #229, opened 2016-06-14
JBDS-3947 fix where vagrant -v output parsing

dgolovin:


dgolovin:
released to master with rebase,

PR #230, opened 2016-06-14
Test: Download 7zip directly from 7zip domain

jrichter1:
Random builds have been failing due to sourceforge redirecting to the respective mirror sometimes taking too long so the requests would time out. Trying out if a direct download from 7-zip.org performs better.


jrichter1:
test this

jrichter1:
test this

PR #235, opened 2016-06-28
Update devstudio URLs

jrichter1:
Build is failing on devstudio link returning 404, updating the appropriate URLs so we can build again.

### DIFF
--- a/requirements.json
+++ b/requirements.json
@@ -44,9 +44,9 @@
     "description": "A certified Eclipse-based integrated development environment (IDE)",
     "bundle": "yes",
     "version": "10.0.0.GA",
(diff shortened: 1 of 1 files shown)


jrichter1:
@dgolovin @ppitonak care for a review?

dgolovin:
I think we should use link to latest build, not specific one, because it is going to be deleted at some point.

jrichter1:
It's the same staged build as before, just the file name changed for some reason, so it probably won't get deleted any time soon.
But if you wish, I'll update it to the latest build - as long as our builds get fixed I'm fine with it.

dgolovin:
I created new one https://github.com/redhat-developer-tooling/developer-platform-install/pull/237

jrichter1:
Well then

PR #236, opened 2016-06-29
Get installables info from requirements.json

dgolovin:


dgolovin:
depends on https://github.com/redhat-developer-tooling/developer-platform-install/pull/237

PR #244, opened 2016-07-13
JBDS-3769, JBDS-3958: JSPM removed, Babel runs at build time

jrichter1:
- JSPM has been removed (at last)
 - Babel updated
 - The entire project is now transpiled using Babel before 'generate'/'run' tasks saving loading time at the start of the app
 - *.es6.js files renamed to *.js since they are no longer transpiled on their own
 - Transpiled sources are put into a new folder, hence model classes needed an update to their bundled files paths


jrichter1:
test this

jrichter1:
Sourceforge seems to be down, can't download 7zip and finish the build...

dgolovin:
test this

dgolovin:
test this

dgolovin:
It works good. The only problem is at development time only transpiled scripts are present and makes debugging difficult.

jrichter1:
I've added sourcemaps to the transpiled scripts, so the original ones should now be visible in the debugger under 'sources'.

dgolovin:
That works, thanks!

PR #251, opened 2016-07-18
Update Patternfly and Protractor

jrichter1:
Looks like the major version bump doesn't break anything for us.


dgolovin:
It seems also required to install bower globally or 
```
npm install
```
fails with installation problems for patternfly.

Should readme.md be updated as well to reflect this?

The npm install also fails in central PR check builds, for example UI Tests has:
```
07:54:08 'bower' is not recognized as an internal or external command,
07:54:08 operable program or batch file.
07:54:08 npm WARN install:bootstrap-treeview@1.2.0 bootstrap-treeview@1.2.0 install: `bower install`
07:54:08 npm WARN install:bootstrap-treeview@1.2.0 Exit status 1
```

dgolovin:
After running 
```
npm install bower -g
```
everything is back to normal in my local builds.

jrichter1:
This is strange, I could just npm install it all and it didn't even need bower in the global modules. I'll take a second look at it.

jrichter1:
You're right about the bower problem. But as far as I can tell, it only serves to install optional dependencies for Patternfly so it won't fail the build when they can't be installed.

We are not using these optional dependencies, the app looks like nothing broke in the UI, so the options are:
 - ignore the warnings, since we don't even need those additional packages
 - install bower, but deal with needless packages
 - keep an older version of Patternfly
My vote is not to install bower.

dgolovin:
Verified again from scratch, still have bower error, but everything runs. 
I agree to not install bower.

dgolovin:
Released as ba475db after re-base.
Thanks Jan!

PR #256, opened 2016-07-22
JBDS-3853 fix sfx metadata to be consistent with electron binary

jrichter1:


dgolovin:
Released to master as abe7690 after rebase.

PR #266, opened 2016-08-04
JBDS-3713 Add a prompt when exiting the installer (i.e. via the cross button)

dgolovin:
Fix adds overall close message handler for all pages except last one.
confirm


jrichter1:
Looks good, except the cancel button on location page is still bound to the old close dialog.

dgolovin:
ups, thanks Jan! I'll fix it.

PR #267, opened 2016-08-05
Save files one by one with progress

dgolovin:


gulpfile.js
```diff
@@ -281,32 +282,31 @@ gulp.task('prefetch-tools', ['create-tools-dir'], function() {
...
     }
   }
-  return Promise.all(promises).then((result) => {
-    if (bundle === 'yes') {
-      installerExe = resolveInstallerExePath('-bundle');
-    }
-  });
-}
+  console.log(promises.length);
+  return promises.reduce( function(pacc, fn) { return pacc.then(fn); }, Promise.resolve() );
```
jrichter1: You'll need to put the installerExe resolution for the bundled installer back in here. This way running 'dist' attempts to produce 2 archives with the same name, so we only end up with one of them.

dgolovin: Thanks! I didn't notice that and was expecting PR build tests that. That is something what should be tested build check, because just running the build is obviously not enough. It should also verify that expected deliverables are produced.

PR #269, opened 2016-08-10
JBDS-3991 create build folder with timestamp

dgolovin:


dgolovin:
no need for unique folder, it created remotely on server and then reg folder dist/win copied over using scp

PR #272, opened 2016-08-10
Update CDK to Aug 10th nightly build

dgolovin:


dgolovin:
released to master after re-base as e162c64bbf6689e9922ddecd1fe5558f1f132641.

PR #276, opened 2016-08-12
Update DevSuite Installer version to 1.1.0-GA

dgolovin:


dgolovin:
Released as  eaaabe5 after re-base

PR #278, opened 2016-08-12
JBDS-3829 - Installer is empty if the user has control char in name

dgolovin:


dgolovin:
Released as c262b34 after re-base.

PR #281, opened 2016-08-15
JBDS-3999 fix openjdk msi/zip confusion

jrichter1:
rename jdk.zip to jdk.msi and update its sha


dgolovin:
Thanks Jan! 
I also noticed that, but by that time you already had this fix.
Released as 815f949 after re-base.

PR #284, opened 2016-08-16
JBDS-3754 Installer: redirect links for Vagrant and VirtualBox should  use https

dgolovin:


requirements.json
```diff
@@ -76,7 +76,7 @@
     "bundle": "no",
     "version": "5.0.26",
     "revision": "108824",
-    "url": "http://download.virtualbox.org/virtualbox/5.0.26/VirtualBox-5.0.26-108824-Win.exe",
+    "url": "http://developers.redhat.com/redirect/to/virtualbox-5.0.26.download",
```
jrichter1: Wasn't this supposed to be https as well?

dgolovin: Original url is http, so it should be http redirect as well.

dgolovin:
Released to master as 5476ad2 after re-base.

PR #288, opened 2016-08-17
Fix cygwin version in requirements.json to be 2.5.2

dgolovin:


dgolovin:
Released to master as fac5d17 after re-base.

PR #291, opened 2016-08-18
JBDS-3834 Install fail - what next?

dgolovin:
This PR adds context menu to show chrome devtools inside installer window.
devtools-popup-menu


dgolovin:
Seems to work as expected, except accelerator. Can be fixed by calling Menu.setAppicationMenu(), but that leads to visible bar menu after Alt key is pressed. It should be fine even w/o hotkey available.

PR #296, opened 2016-08-20
JBDS-4005 Use https redirects for cygwin-installer and oc origin cli

dgolovin:


dgolovin:
test this

dgolovin:
Released as 3cba3a1 after re-base.

PR #297, opened 2016-08-22
Make progress bars useful again

jrichter1:
An attempt to make the progress bars more friendly than they appear to be now:
 - determinate progress only for downloads, now includes downloaded / total sizes
 - waiting and installing switches to animated indeterminate progress
 - failed components' progress bar turns red
 - completed progress bar turns green
 - restarting downloads resets the downloading progress

Comments, criticism, issues, etc. welcome

### DIFF
--- a/browser/directives/progressBar.js
+++ b/browser/directives/progressBar.js
@@ -27,7 +27,8 @@ function progressBar() {
               <div>{{productDesc}}</div>
             </div>
             <div class="progress progress-label-top-right">
-              <div class="progress-bar" role="progressbar" aria-valuenow="{{current}}" aria-valuemin="{{min}}" aria-valuemax="{{max}}" style="width: {{current}}%;">
+              <div class="progress-bar" ng-class="{'progress-bar-striped active': status == 'Installing' || status == 'Setting up' || status.indexOf('Waiting') > -1}" role="progressbar" aria-valuenow="{{current}}" aria-valuemin="{{min}}" aria-valuemax="{{max}}"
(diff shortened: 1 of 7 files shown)


jrichter1:
progress

dgolovin:
Does it handle re-download of one file from CDK? It seems just reset counters to 0, but it should decrease downloaded by actual amount downloaded before failure.

What if progress is shown only for setup, when vagrant and virtualbox were detected?

dgolovin:
From design point of view I think we should keep it as close as possible to document attached here https://patternfly.atlassian.net/browse/RHUAP-26. There is no colored progress bars in design UX team suggested.

dgolovin:
Could you separate progress for download,  waiting/installing switches to animated indeterminate progress and coloring failed/completed progress and submit them as separated PR's?

jrichter1:
A quick answer compilation:
 - restarting doesn't work on a per file basis - will definitely look into that
 - setup bars look the same as installation ones
 - colouring can be omitted / separated from the rest no problem
 - splitting download/wait/install etc. could make a mess of things

jrichter1:
Fixed up the damn error - now it should correctly add up the download size for multiple files, plus restarting download now works properly if some file already finished downloading.

Also removed the colouring for now, so it's all blue (downloading, finished) or striped blue (waiting, installing, setting up).

dgolovin:
That is awesome! Thanks! Just in time for next PR with download-manager links for testing.
Released to master as  707a1e2 after re-base.

PR #302, opened 2016-08-30
JBDS-3990 Create mock server to test online installer

dgolovin:
Demo https://youtu.be/L41p1_tlvo8

1. http/https servers implemented. both required, because we have
http and https redirects in requirements.json;
2. account/controller.js using request module to check T&C's instead of
angular http services, to be able to force ignore self signed sertificate
for mock https server;
3. self signed sertificate error ignored by setting env variable
DSI_REJECT_UNAUTHORIZED=false;
4. developers.redhat.com mapped to localhost in hosts file in
c:\Windows\system32\drivers\etc\hosts;
5. gulp prefetch-all task added to download all binaries
before running mock serveres;
6. Run servers using node gulp-tasks/http(s)-server.js


dgolovin:
@jrichter1 This PR has test failures in unit-tests, it seems long task to mock all angular services. Could we just relay on gulp ui-test? Would that work for QA?

ppitonak:
I'm missing documentation in this PR, either in README or another file.

jrichter1:
@dgolovin Unit tests failing? No, sir, that would not work at all (quite a mess that would make). You know the drill - you change the implementation, you change the tests accordingly.

jrichter1:
How about something like this?
```javascript
timeout = function(cb) { cb(); }
scope = { $apply: function(cb) { cb(); } };
```

dgolovin:
@jrichter1 that code covered in ui-test anyway, so my proposal is not skip to tests at all, but have ui tests only. But I'll see what I can do in next couple days.

dgolovin:
@ppitonak will update README.md.

dgolovin:
@jrichter1 thanks for mocking ideas, tests are fixed :)
@ppitonak README.md updated.

PR #304, opened 2016-08-30
JBDS-3989 Add projected public links for all components

dgolovin:


dgolovin:
replaced with https://github.com/redhat-developer-tooling/developer-platform-install/pull/316

PR #307, opened 2016-08-31
JBDS-4026 Bundled binaries sha256 is not verified after download

dgolovin:


dgolovin:
test this

gulpfile.js
```diff
@@ -151,23 +151,49 @@ gulp.task('create-7zip-archive', function(cb) {
...
+          reqs['jbds.jar'].sha256sum = body;
+          resolve();
+        });
       }
-    }
-  });
+    });
+  };
+
+  return Promise.resolve()
```
jrichter1: Is this line really necessary?

dgolovin: This line triggers Promises chain. It just looks better than just calling first function to start the chain.

jrichter1: Well, if you look at the model classes, the promise chains usually start with the first business function. I'm not a fan of this padded construction (but it's no big deal either).

PR #308, opened 2016-09-01
JBDS-4031 error sha check for downloaded cygwin

dgolovin:
Always include cygwin.exe in installer for one big reason. If file gets updated to new one, there is no archived copy of previous one and that means redirect we use does not help.


dgolovin:
test this

dgolovin:
test this

PR #310, opened 2016-09-01
JBDS-4043 Update cygwin version to 2.6.0

dgolovin:
Fix updates version shown in installer to 2.6.0 and place direct link
instead of redirect because cygwin installer (size under 1M) will be
always included in online and bunded installer after
https://issues.jboss.org/browse/JBDS-4031 is fixed

It means it is  going to be alway downloaded during the build.


dgolovin:
test this

PR #311, opened 2016-09-01
JBDS-4036 Installer should verify detected Virtualbox is from 5.0.x stream

dgolovin:


jrichter1:
Maybe a new error message would be in order as well?

jrichter1:
So, 5.1 will no longer be valid, right? That calls for a new error message.

PR #312, opened 2016-09-02
JBDS-4041 cygwin installer fails to start with EBUSY error in install.log

dgolovin:
Fix changes event listeners to 'close' and that fixes the issue.


dgolovin:
test this

PR #313, opened 2016-09-02
JBDS-4007 Need a new check for existing OpenJDK installed as MSI

dgolovin:
Fix implements:
1. Check if OpenJDK MSI package installed, using the same query
   used by uninstall.ps1. It sets flag openJdkMsi on JdkInstall
   instance. This flag blocks OpenJDK selection for installation
   to avoid possible installation errors in case msi cannot update
   already installed OpenJDK;
2. Disabling for Back ad Next buttons on Confirmation page until
   detection is finished.

### DIFF
--- a/browser/model/helpers/util.js
+++ b/browser/model/helpers/util.js
@@ -21,7 +21,7 @@ class Util {
     });
   }
 
-  static executeFile(file, args, outputCode) {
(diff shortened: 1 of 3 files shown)


dgolovin:
test this

PR #323, opened 2016-09-28
JBDS-4047 uninstaller does not clean up path

dgolovin:
Fix corrects target folder path being used for
filtering Path variable entries. It changes root folder
to be one level up from uninstaller folder.


dgolovin:
test this

PR #332, opened 2016-10-04
Update hardcoded dist/win folder to be based on current platform/arch

dgolovin:
Fix removes unused local var and updates rootbuildfolder to be
named after current platform-arch, rather than hardcoded to
dist/win. Tis fix after merge will require to update nightly
build publishing config on central CI.

### DIFF
--- a/protractor-conf.js
+++ b/protractor-conf.js
@@ -1,5 +1,6 @@
 var files;
 var report;
+var platform =  process.platform + '-' + process.arch;
 
 if (process.env.PTOR_TEST_RUN === 'system') {
   files = ['test/system/*.js'];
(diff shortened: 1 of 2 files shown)


protractor-conf.js
```diff
@@ -20,7 +20,7 @@ exports.config = {
   capabilities: {
     browserName: 'chrome',
     chromeOptions: {
-      binary: './dist/win/devsuite-win32-x64/devsuite.exe'
+      binary: './dist/' + process.platform + '-' + process.arch +'/devsuite-win32-x64/devsuite.exe'
```
jrichter1: For consistency's sake, I suggest replacing the 'devsuite-win32-x64' part as well.

dgolovin: Thank Jan! I updated protractor configuration.

PR #337, opened 2016-10-06
JBDS-4090 Virtualbox detection fails on Mac OS X

dgolovin:


dgolovin:
released to master as 6594ac4 after rebase.

PR #339, opened 2016-10-07
JBDS-4092 WIP Mac OS X devstudio.app starts with empty screen

dgolovin:
Fix gets rid of resolution where req/package.json located.
It explicitly deefines node root location and adjust imort
statements thoughout the tests. Plus adds configuration for
protractor targeted installer executable that depends on
plaform ui-tests are running. Also it updates reuire calls
with packjage.json and requirements.json.

Tested on windows and mac os x. ui-tests have many errors
under Mac, but it is not a concern for this fix.


dgolovin:
test this

PR #344, opened 2016-10-11
Move requirements.json loading to InstallerDataService

dgolovin:
Fix makes InstallerDataService instanse the only provider of requirements information.


dgolovin:
Released to master as ba5b617 after re-base.

PR #345, opened 2016-10-12
JBDS-4099 DevStudio does not start from installer on Mac

dgolovin:


dgolovin:
Released to master as dbdadbf after re-base.

PR #353, opened 2016-10-14
JBDS-3875 Split gulpfile.js to multiple modules to simplify maintenance

dgolovin:
Fix moves out functions to different files to be able to share them
between different guil related files.


dgolovin:
test this

PR #354, opened 2016-10-15
JBDS-4080  Avoid to redownload everything on each install

dgolovin:
Fix adds verification for downloaded files in temp folder and
if sha is the same installer uses those files instead of downloading
them again.

### DIFF
--- a/browser/model/installable-item.js
+++ b/browser/model/installable-item.js
@@ -1,12 +1,16 @@
 'use strict';
 
 import Util from './helpers/util';
+import Hash from './helpers/hash';
+import Logger from '../services/logger';
 import path from 'path';
+import fs from 'fs-extra';
+import Downloader from './helpers/downloader';
 
 let ipcRenderer = require('electron').ipcRenderer;
 
 class InstallableItem {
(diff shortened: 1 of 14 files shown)


browser/model/installable-item.js
```diff
@@ -113,6 +116,38 @@ class InstallableItem {
     success();
   }
 
+  checkAndDownload(dl,url,sha,user,pass) {
```
jrichter1: I believe we should provide unit test with new methods. Please add tests for this method. And if you could remove any unnecessary debug messages, that would be smashing.

browser/model/installable-item.js
```diff
@@ -113,6 +116,38 @@ class InstallableItem {
...
+          this.startDownload(dl,url,sha,user,pass);
+        }
+      });
+    } else {
+      console.log("here1------------------------------------------------------------------------");
+      this.startDownload(dl,url,sha,user,pass);
+    }
+  }
+
+  startDownload(dl,url,sha,user,pass) {
```
jrichter1: Please provide unit tests as well for this method.

dgolovin:
Refactored download a little bit to get rid of code duplication and added tests for new methods.

jrichter1:
Chick Norris Approves

PR #358, opened 2016-10-18
Update DevStudio to latest 10.2.0.AM2 respin-y

dgolovin:


nickboldt:
There is no respin-y. I was only experimenting with publishing scripts. Use AM2, not AM2*

dgolovin:
NP, Nick! I'll send one more PR for latest staging build. Thanks!

dgolovin:
We actually all good with what we have in requirements.json now :)

PR #372, opened 2016-10-25
JBDS-4139 undefined key used in cdk whenlogging PATH env var value

dgolovin:
Fix updates the code to use static method to get the key instead of this.key.


dgolovin:
test this

PR #373, opened 2016-10-26
JBDS-4140 env['Path'] doesn't work for child_process on OS X

dgolovin:
Fix updates name to 'PATH" because it works for OS X and Window


dgolovin:
test this

PR #381, opened 2016-11-01
Increase unit-tests coverage for JdkInstall to 100%

dgolovin:
Fix adds missing tests to increase coverage to 100%

### DIFF
--- a/test/unit/model/jdk-install-test.js
+++ b/test/unit/model/jdk-install-test.js
@@ -7,10 +7,10 @@ import mockfs from 'mock-fs';
 import request from 'request';
 import fs from 'fs-extra';
 import path from 'path';
-import JdkInstall from 'browser/model/jdk-install';
 import Logger from 'browser/services/logger';
 import Downloader from 'browser/model/helpers/downloader';
 import Util from 'browser/model/helpers/util';
+import JdkInstall from 'browser/model/jdk-install';
 import Installer from 'browser/model/helpers/installer';
 import Hash from 'browser/model/helpers/hash';
 import InstallableItem from 'browser/model/installable-item';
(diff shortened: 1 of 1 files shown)


dgolovin:
test this

test/unit/model/jdk-install-test.js
```diff
@@ -162,7 +162,26 @@ describe('JDK installer', function() {
...
+      }
+    });
 
+    it('should check for available msi installtion', function(done) {
+      mockDetectedJvm('1.8.0_1');
+      let jdk = new JdkInstall(installerDataSvc, 'url', 'file');
+      jdk.findMsiInstalledJava.restore();
+      return jdk.detectExistingInstall(function() {
+        expect(executeFileStub).to.have.been.calledOnce;
+        expect(writeFileStub).to.have.been.calledOnce;
```
jrichter1: I've looked into the test failures on CI and found this particular test timed out because the assertions are not met and the eventual error is consumed by the callback.

The root of the problem is on this line: https://github.com/redhat-developer-tooling/developer-platform-install/blob/master/browser/model/jdk-install.js#L102
The tests are being run on a RHEL machine and these methods are only run on Windows. In the spirit of the unit tests running on more platforms, I'd request this piece of code be executed for Windows only.

dgolovin: Thanks @jrichter1, I was thinking about ether skip it for other platforms or override in some way process.platform and let run on stabbed functions.

dgolovin:
test this

dgolovin:
test this

test/unit/model/jdk-install-test.js
dgolovin: I have updated updated PR. Now it checks if windows specific code is called and worked as expected, else branch does the opposite and verifies windows specific code is not called on other platforms.

dgolovin:
test this

dgolovin:
test this

dgolovin:
@jrichter1, I've fixed test error on linux, I am going to release this PR in master.

dgolovin:
thanks @jrichter1 !

PR #392, opened 2016-11-07
JBDS-4163 undefined have no valid property exception

dgolovin:
Fix adds:
* default implementation for validateVersion method
* change var declaration to let in cycle to def validation callbacks
* add check for selected option to be present


dgolovin:
test this

PR #393, opened 2016-11-07
JBDS-4119 Uninstaller - remove files with too long path

jrichter1:
In case some files cant get deleted due to too long path, use robocopy (available in Win 7+) to purge the directories.


dgolovin:
released to master, thanks!

PR #394, opened 2016-11-07
JBDS-4028 Use direct link to registration form

jrichter1:
So it actually gets you to registration instead of login


dgolovin:
Released to master, thanks!

PR #399, opened 2016-11-08
Update DevStudio build to latest nightly build

dgolovin:
Fix includes direct links to avoid downloading after
new nightly version is published during demo


dgolovin:
test this

PR #401, opened 2016-11-08
JBDS-4165 Version parsing error

dgolovin:
Fix parce major, minor and pathc versions from full version to
use with semver module instead of full version with underscore


dgolovin:
test this

dgolovin:
test this

PR #412, opened 2016-11-18
Improve code coverage for JDK Installer

dgolovin:
Fix adds platform specific tests to JDK unit tests,
corrects wrong expect calls and gets coverage to
back to 100%.


dgolovin:
Test this

PR #431, opened 2016-12-01
Codecov dependency and script added

ppitonak:
See manually imported results in https://codecov.io/gh/redhat-developer-tooling/developer-platform-install. I will update Jenkins jobs after this PR is merged, otherwise it would cause them to fail (non-existing npm script).

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,13 +15,14 @@
   },
   "homepage": "https://github.com/redhat-developer-tooling/developer-platform-install#readme",
   "directories": {
-    "app" : "transpiled"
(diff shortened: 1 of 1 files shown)


ppitonak:
test this

dgolovin:
@ppitonak is that possible to report coverage for PR on github the same way it is done for almighty-core PR's?

dgolovin:
👍 it looks good

ppitonak:
Thanks for merging. It's configured on Jenkins now so all new PRs will receive a comment from codecov.

PR #445, opened 2016-12-08
JBDS-4170 Remove windows uninstaller from mac distribution

sudhirverma:
Remove windows uninstaller from mac distribution

### DIFF
--- a/test/unit/services/data-test.js
+++ b/test/unit/services/data-test.js
@@ -6,6 +6,7 @@ import { default as sinonChai } from 'sinon-chai';
 import VagrantInstall from 'browser/model/vagrant';
 import VirtualBoxInstall from 'browser/model/virtualbox';
 import InstallerDataService from 'browser/services/data';
+import Platform from 'browser/services/platform';
 import Logger from 'browser/services/logger';
 import path from 'path';
 import os from 'os';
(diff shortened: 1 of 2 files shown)


test/unit/services/data-test.js
```diff
@@ -106,6 +107,18 @@ describe('InstallerDataService', function() {
...
   });
 
+  describe('copyUninstaller', function(){
+    it('should be able to handle copyUninstaller', function(){
+      let data = {
+          darwin: ()=>'darwin',
+      };
+      sandbox.stub(Platform,'getOS').returns('darwin');
+      expect(Platform.identify(data)).to.be.equal('darwin');
+      sandbox.spy(svc, 'copyUninstaller');
```
dgolovin: You need to call setup method here
``` JavaScript
svc.setup();
```
and this test will fail, because issue itself is not fixed yet.

test/unit/services/data-test.js
```diff
@@ -106,6 +107,18 @@ describe('InstallerDataService', function() {
     });
   });
 
+  describe('copyUninstaller', function(){
+    it('should be able to handle copyUninstaller', function(){
+      let data = {
+          darwin: ()=>'darwin',
+      };
+      sandbox.stub(Platform,'getOS').returns('darwin');
+      expect(Platform.identify(data)).to.be.equal('darwin');
```
dgolovin: You should remove lines 112-114 and 116, because they do not needed here. Line 115 is enough to make installer thing it is running on macOS.

dgolovin: You should remove lines 112-114 and 116, because they do not needed here. Line 115 is enough to make installer think it is running on macOS.

test/unit/services/data-test.js
dgolovin: After all above is fixed you can fix the issue here https://github.com/redhat-developer-tooling/developer-platform-install/blob/master/browser/services/data.js#L53. Fix is really simple
``` JavaScript
if(Platform.OS === 'win32') {
    this.copyUninstaller();
}
```
This change will fix unit test failure.

dgolovin:
@sudhirverma LGTM, you now can mark [JBDS-4170](https://issues.jboss.org/browse/JBDS-4170) as resoved.

PR #449, opened 2016-12-12
Extend system tests

jrichter1:
- add tests for user defined target folder
 - add tests for component detection
 - scripts for CI updated
 - add some ids to elements for lookup


dgolovin:
@jrichter1 we need to put CI shell scripts in SCM as well or even whole jenkins configuration. It would be probably private repository or internal gitlab instance. Otherwise changes to CI are not visible.
WDYT?

jrichter1:
@dgolovin I don't really mind either way. 
As far as the scripts go, they can still be used as a convenience method to run the tests, so there is no harm to have them here. They don't contain anything internal.
But I agree about the configuration, we should probably store it somewhere. Do you happen to have a nice repo ready for such a thing?

dgolovin:
Nope, I have nothing ready for this. I know devstudio has private repo for this on github, but I would use gitlab instance.

dgolovin:
@jrichter1 I am +1 to release it to master.  We'll take care of CI shell scripts sharing later.

jrichter1:
Very well, I shall get to it once I figure out what the hell I was doing before holidays kicked in :)

jrichter1:
Released to master as 62f2940.

PR #451, opened 2016-12-13
JBDS-4130 Environment variables are duplicated after several installations

sudhirverma:
Environment variables are duplicated after several installations


jrichter1:
Though a nice change, this does not actually solve the linked issue. The point of the issue was not the code duplication, but rather the duplication of environment variables themselves (i.e. we should check if the variable already exists before adding it).

browser/services/platform.js
```diff
@@ -1,6 +1,16 @@
 'use strict';
 
 class Platform {
+
+  data(installerDataSvc){
```
jrichter1: I think you can safely make this static so you don't need to instantiate the class every time just for the sake of calling the method.

PR #454, opened 2016-12-13
Fix promise rejection error in cygwin tests

dgolovin:
Fix removes done() call to fix the error.

### DIFF
--- a/test/unit/model/cygwin-test.js
+++ b/test/unit/model/cygwin-test.js
@@ -231,8 +231,8 @@ describe('Cygwin installer', function() {
         expect(installer.hasOption('detected')).to.be.equal(true);
       });
     });
-    describe('on Windows', function(done) {
-      it('should mark cygwin as detected when cygwin, openssh and rsync pacakes are installed', function() {
+    describe('on Windows', function() {
+      it('should mark cygwin as detected when cygwin, openssh and rsync pacakes are installed', function(done) {
         sandbox.stub(Platform, 'getOS').returns('win32');
         sandbox.stub(Util, 'executeCommand').onFirstCall().returns(Promise.resolve(
           ['Cygwin Package Information',


dgolovin:
test this

jrichter1:
I don't think this change is going to fix anything. The problem here is that the callback in detectExistingIstall(cb) is in fact swallowing assertion errors. So if the assertion actually fails it only puts the Unhandled promise rejection error into output, but the test still passes.
You can go ahead and try it with some bogus expectations. It just won't make the test fail.
This needs a different approach, maybe we could get rid of the callback and just chain the promise forward?

test/unit/model/cygwin-test.js
```diff
@@ -245,7 +245,6 @@ describe('Cygwin installer', function() {
         installer.detectExistingInstall(function() {
```
jrichter1: This is the call that needs a return statement to be executed properly. I guess there are other in the code as well. Can you take a look?

dgolovin:
if mocha callback executed at the end of detection callback everything seems to be working properly.

dgolovin:
I updated PR with right fix and tested with wrong expectations and it fails, becuase done method is not being called at the end of callback.

jrichter1:
Actually you can remove the done callback/function. The problem was that in order to test an async function you need to call it with return. As in

return asyncFunction(callback() {
  expect(something)
})

dgolovin:
This should work for promises, but obviously does not for what we have now.

jrichter1:
Right, silly me. Is there a particular reason to have the promise chain end with a callback that then get wrapped into a promise again?

dgolovin:
Do you mean pomisify detectExistingInstall()?

dgolovin:
test this

jrichter1:
Yes, technically it should just be about replacing the callback with a resolve/reject and removing the wrapper in confirm controller. (And changing the tests.)

dgolovin:
I'll refactor that. Should we push this one? I will file a jira for refactoring.

dgolovin:
test this

dgolovin:
Opened https://issues.jboss.org/browse/JBDS-4223 to refactor detection merhods to return promise.

PR #458, opened 2016-12-16
JBDS-4224 Update OpenJDK to latest available release

dgolovin:
Fix updates public and internal links for OpenJDK release
and adds extra parameter to msiexec to install jdk and updater.


dgolovin:
test this

PR #473, opened 2016-12-24
Add tests for files missing in coverage report

dgolovin:


dgolovin:
Obviously, there is bettera  way to implement this. Instead of creating new mediator between electron and application, controllers should get Electron instance through AngularJS DI. Then in unit tests it would be much easier to achieve higher coverage by injecting fake electron implementation using Sinon.JS mocks/stubs.

browser/services/electron.js
```diff
@@ -0,0 +1,31 @@
+'use strict';
```
dgolovin: Obviously, there is bettera way to implement this. Instead of creating new mediator between electron and application, controllers should get Electron instance through AngularJS DI. Then in unit tests it would be much easier to achieve higher coverage by injecting fake electron implementation using Sinon.JS mocks/stubs.

dgolovin:
test this

dgolovin:
releasing to have code coverage include uncovered angular controller instances.  Tests are going to be marked as pending.

PR #475, opened 2016-12-29
Increase unit tests code coverage installable item

sudhirverma:
Increase unit tests code coverage installable item

### DIFF
--- a/test/unit/model/installable-item-test.js
+++ b/test/unit/model/installable-item-test.js
@@ -14,9 +14,11 @@ chai.use(sinonChai);
 
 describe('InstallableItem', function() {
 
-  let infoStub;
+  let infoStub, item;
   before(function() {
     infoStub = sinon.stub(Logger, 'info');
+    let svc = new InstallerDataService();
+    item = new InstallableItem('jdk', 1000, 'url', 'installFile', 'targetFolderName', svc);
   });
 
   after(function() {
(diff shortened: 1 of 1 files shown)


test/unit/model/installable-item-test.js
```diff
@@ -135,4 +135,53 @@ describe('InstallableItem', function() {
 
   });
 
+  describe('InstallableItem test', function(){
+    let svc = new InstallerDataService();
```
dgolovin: This line and next one should be wrapped into before() method
``` JavaScript
let item;
before(function(){
  let svc = new InstallerDataService();
  item = new InstallableItem('jdk', 1000, 'url', 'installFile', 'targetFolderName', svc);
});
```

test/unit/model/installable-item-test.js
```diff
@@ -135,4 +135,53 @@ describe('InstallableItem', function() {
...
+    let item = new InstallableItem('jdk', 1000, 'url', 'installFile', 'targetFolderName', svc);
+    it('get install time', function() {
+      item.getInstallTime();
```
dgolovin: There is no reason to call method twice during the test here and in tests below

test/unit/model/installable-item-test.js
```diff
@@ -135,4 +137,43 @@ describe('InstallableItem', function() {
...
+
+    it('get install time', function() {
+      expect(item.getInstallTime()).to.be.equal(1000);
+    });
+
+    it('has exi', function() {
+      expect(item.hasExistingInstall()).to.be.equal(false);
+    });
+
+    it('check for existing install', function() {
```
dgolovin: This test should follow the pattern for functional test

``` JavaScript
describe('checkForExistingInstall abstract method', function(){
   it('should not add detected option when called',function(){
       // implement test here
   });
});

```

PR #476, opened 2016-12-30
Increase unit tests code coverage for downloader test

sudhirverma:
Increase unit tests code coverage for downloader test

### DIFF
--- a/test/unit/model/helpers/downloader-test.js
+++ b/test/unit/model/helpers/downloader-test.js
@@ -8,6 +8,7 @@ import Downloader from 'browser/model/helpers/downloader';
 import Logger from 'browser/services/logger';
 import { Readable, PassThrough, Writable } from 'stream';
 import Hash from 'browser/model/helpers/hash';
+import fs from 'fs-extra';
 chai.use(sinonChai);
 
 describe('Downloader', function() {
@@ -40,6 +41,7 @@ describe('Downloader', function() {
 
   beforeEach(function() {
     sandbox = sinon.sandbox.create();
+    downloader = new Downloader(fakeProgress, function() {}, function() {});
   });
 
   afterEach(function() {
@@ -47,12 +49,10 @@ describe('Downloader', function() {
   });
 
   it('should set totalDownloads to 1 by default', function() {
-    downloader = new Downloader(fakeProgress, function() {}, function() {});
     expect(downloader.totalDownloads).to.equal(1);
   });
 
   it('responseHandler should set the total download size', function() {
-    downloader = new Downloader(fakeProgress, function() {}, function() {});
     let response = { headers: { 'content-length': 1024 } };
 
     downloader.responseHandler(response);
@@ -61,7 +61,6 @@ describe('Downloader', function() {
   });
 
   it('dataHandler should update the progress once time threshold is reached', function() {
-    downloader = new Downloader(fakeProgress, function() {}, function() {});
     fakeProgress.totalSize = 1024;
 
     let data = { length: 512 };
@@ -75,7 +74,6 @@ describe('Downloader', function() {
   });
 
   it('dataHandler should not update the progress before time threshold is reached', function() {
-    downloader = new Downloader(fakeProgress, function() {}, function() {});
     fakeProgress.totalSize = 1024;
 
     let data = { length: 512 };
@@ -105,8 +103,6 @@ describe('Downloader', function() {
   it('endHandler should end the stream', function() {
     let stream = { end: function() {} };
     let streamSpy = sandbox.spy(stream, 'end');
-
-    downloader = new Downloader(fakeProgress, function() {}, function() {});
     downloader.setWriteStream(stream);
     downloader.endHandler(stream);
 
@@ -114,7 +110,6 @@ describe('Downloader', function() {
   });
 
   it('closeHandler should verify downloaded files checksum', function() {
-    downloader = new Downloader(fakeProgress, function() {}, function() {});
     let stub = sandbox.stub(Hash.prototype, 'SHA256').yields('hash');
 
     downloader.closeHandler('file', 'hash', 'url');
@@ -154,7 +149,6 @@ describe('Downloader', function() {
 
     it('should make a request with given options', function() {
       let requestGetSpy = sandbox.spy(request, 'get');
-      downloader = new Downloader(fakeProgress, function() {}, function() {});
       downloader.setWriteStream(new PassThrough());
       downloader.download(options3);
 
@@ -164,7 +158,6 @@ describe('Downloader', function() {
 
     it('should make a request with given url in options', function() {
       let requestGetSpy = sandbox.spy(request, 'get');
-      downloader = new Downloader(fakeProgress, function() {}, function() {});
       downloader.setWriteStream(new PassThrough());
       downloader.download(options);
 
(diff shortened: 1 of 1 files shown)


test/unit/model/helpers/downloader-test.js
```diff
@@ -270,5 +270,15 @@ describe('Downloader', function() {
 
       expect(successHandler).to.be.calledOnce;
     });
+
+    it('should able to test restartDownload and downloadAuth', function() {
```
dgolovin: You targeting restartDownload() function is this test. Then it should not be inside `describe('download', function(){` instead you should create 
``` JavaScript
describe('restartDownload', function() {
});
```
as a sibling for it to follow the pattern
* Class
  * method
     * should test 1
     * should test 2

``` JavaScript
...
describe('download', function() {
...
     it('should make a request with given options', function() {
...
     });
...
     it('should make a request with given url in options', function() {
...
     });
});

describe('restartDownload', function() {
    // create pending tests
});
```

then fill it with pending tests that would reads as specification for the function

``` JavaScript
describe('restartDownload', function() {
    it('should change downloader status from \'Download Failed\' to \'Downloading\'');
    it('should create new WriteStream instance for failed download entries');
    it('should call authDownload for entries that requires authentication');
    it('should call download method for entries that does not require authentication');
});
```

when you run unit tests you would see results report matching pattern defined above.
``` text
 Downloader
    √ should set totalDownloads to 1 by default
    √ responseHandler should set the total download size
    √ dataHandler should update the progress once time threshold is reached
    √ dataHandler should not update the progress before time threshold is reached
    √ errorHandler should close the stream
    √ endHandler should end the stream
    √ closeHandler should verify downloaded files checksum
    √ closeHandler should call success when verification succeeds
    √ closeHandler should call failure when verification fails
    download
      √ should make a request with given options
      √ should make a request with given url in options
      √ should make a request with 'User-Agent' header set
      √ should call endHandler when end event is emitted
      √ should call errorHandler when error event is emitted
      √ should save downloads in map
      √ should not call sucessHandler after error event is emitted
      √ should call sucessHandler ony after all downloads are finished
    restartDownload
      - should change downloader status from 'Download Failed' to 'Downloading'
      - should create new WriteStream instance for failed download entries
      - should call authDownload for entries that requires authentication
      - should call download method for entries that does not require authentication
```

then you implement every pending test to pass.

test/unit/model/helpers/downloader-test.js
```diff
@@ -271,4 +272,31 @@ describe('Downloader', function() {
       expect(successHandler).to.be.calledOnce;
     });
   });
+
+  describe('restartDownload', function(){
+    let options = 'http://example.com/jdk.zip';
+
+    it('should change downloader status from \'Download Failed\' to \'Downloading\'', function(){
+      expect(downloader.restartDownload()).to.be.equal(undefined);
```
dgolovin: restartDownload() function does not have 'return' statement in it. This means the function always returns 'undefined' unless exception occurs. There is no need to have this expect call here. 

In test you should create downloader instance first and simulate download error. See example on [lines 205 - 215](https://github.com/redhat-developer-tooling/developer-platform-install/pull/476/files#diff-762967bfd2ff9393807ffee1c812a190R205) except you don't need to stub errorHandler method.
After that downloader will have downloads Map instance populated with url for which download was failed.

Status change for downloader is should be done through progress.setStatus('Downloading') call, so you need to add a spy for fakeProgress object. 

After all preparations above you ready to call the method under test and verify that 'downloader.downloadSize', 'downloader.received', 'downloader.currentSize' were changed to 0 and setStatus spy called with 'Downloading'.

test/unit/model/helpers/downloader-test.js
```diff
@@ -271,4 +272,31 @@ describe('Downloader', function() {
...
+    });
+
+    it('should create new WriteStream instance for failed download entries', function(){
```
dgolovin: This test is not required, this should be covered by next two tests.

test/unit/model/helpers/downloader-test.js
```diff
@@ -271,4 +272,31 @@ describe('Downloader', function() {
...
+
+      let spy = sandbox.spy(fs, 'createReadStream');
+      downloader.restartDownload();
+      expect(spy).not.called;
+    });
+
+    it('should call authDownload for entries that requires authentication', function(){
+      let donwloadauth = downloader.downloadAuth(options, 'username', 'password', 'key', 'sha');
```
dgolovin: Every test inside `describe('restartDownload')` is supposed to have similar structure. 
Each test would have:
* initialization part where you prepare for testing specific invariant/scenario
* `downloader.restartDownload()` call 
* verification part where it would check all expectations

This test and next one should test restartDownload method and verify that it behaves as expected for different kinds of downloaded resources. We have two kinds of resources here one that requires authentication and one that does not.

you can use the same [lines 205 - 215](https://github.com/redhat-developer-tooling/developer-platform-install/pull/476/files#diff-762967bfd2ff9393807ffee1c812a190R205) code as example for initialization part. Bot not just copy it as is :), add required changes to control restartDownload() behavior.

test/unit/model/helpers/downloader-test.js
```diff
@@ -271,4 +258,40 @@ describe('Downloader', function() {
       expect(successHandler).to.be.calledOnce;
     });
   });
+
+  describe('restartDownload', function(){
+    let options = 'http://example.com/jdk.zip';
+
+    it('should change downloader status from \'Download Failed\' to \'Downloading\'', function(){
```
dgolovin: This is good one 👍

test/unit/model/helpers/downloader-test.js
```diff
@@ -271,4 +258,40 @@ describe('Downloader', function() {
...
+      expect(downloader.currentSize  ).to.be.equal(0);
+      expect(fakeProgress.setStatus).to.have.been.calledOnce;
+      expect(fakeProgress.setStatus).to.have.been.calledWith('Downloading');
+    });
+
+    it('should call authDownload for entries that requires authentication', function(){
+      let requestGetSpy = sandbox.spy(request, 'get');
+      downloader.setWriteStream(new PassThrough());
+      downloader.restartDownload();
+      downloader.downloadAuth(options, 'username', 'password', 'key', 'sha');
```
dgolovin: You are almost there :)
You need to add a little more initialization code before calling `restartDownload()`
``` JavaScript
  let response = new Readable();
  sandbox.stub(request, 'get').returns(response);
  let error = new Error('something bad happened');
  let stream = new Writable();
  downloader.setWriteStream(stream);
  downloader.downloadAuth(options);
  response.emit('error', error);
  // after this downloader will have one failed download in downloads map
  sandbox.stub(downloader, 'downloadAuth');  
  downloader.restartDownload();
  expect(downloader.downloadAuth).to.be.calledOnce;
  expect(downloader.downloadAuth).to.be.calledWith(options, 'username', 'password', 'key', 'sha');
```

test/unit/model/helpers/downloader-test.js
```diff
@@ -271,4 +258,40 @@ describe('Downloader', function() {
...
+      expect(requestGetSpy).to.be.calledOnce;
+      expect(requestGetSpy.args[0][0].url).to.be.equal(options, 'username', 'password', 'key', 'sha');
+
+    });
+
+    it('should call download method for entries that does not require authentication',function(){
```
dgolovin: Basically it should be the same as test for `downloadAuth`, but use `download` method instead.

test/unit/model/helpers/downloader-test.js
```diff
@@ -40,19 +40,18 @@ describe('Downloader', function() {
 
   beforeEach(function() {
     sandbox = sinon.sandbox.create();
+    downloader = new Downloader(fakeProgress, function() {}, function() {});
```
dgolovin: That is good move 👍

PR #482, opened 2017-01-09
Remove debug error from cdk watcher

jrichter1:
I couldn't find any practical use for it and it just clutters the console.


dgolovin:
My bad, thanks Jan!

PR #501, opened 2017-01-14
Increase unit tests code coverage controller

sudhirverma:
Increase unit tests code coverage controller

### DIFF
--- a/test/unit/pages/confirm/controller-test.js
+++ b/test/unit/pages/confirm/controller-test.js
@@ -1,31 +1,87 @@
 'use strict';
-require('../../../angular-test-helper');
-require('browser/main');
 
 import chai, { expect } from 'chai';
 import sinon from 'sinon';
 import { default as sinonChai } from 'sinon-chai';
 import 'sinon-as-promised';
+import Logger from 'browser/services/logger';
+import ElectronMock from '../../../mock/electron';
+import InstallerDataService from 'browser/services/data';
+import InstallableItem from 'browser/model/installable-item';
+import ConfirmController from 'browser/pages/confirm/controller';
+
+require('../../../angular-test-helper');
+require('browser/main');
+
 chai.use(sinonChai);
 
 describe('ConfirmController', function() {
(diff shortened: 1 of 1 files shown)


jrichter1:
test this

test/unit/pages/confirm/controller-test.js
```diff
@@ -1,31 +1,93 @@
...
+  describe('exit', function() {
+    it('exit closes active window', function() {
+      sandbox.stub(electron.remote.currentWindow);
+      confirmcontroller.exit();
+      expect(electron.remote.currentWindow.close).calledOnce;
     });
+  });
 
+  describe('download', function() {
+    it('should able to check download', function() {
```
dgolovin: Better name it 'should open url in browser using electron.shell.openExternal' and use test/mock/electron.js in test.

test/unit/pages/confirm/controller-test.js
```diff
@@ -1,31 +1,93 @@
...
+      expect(confirmcontroller.download('url')).to.be.equal(undefined);
```
dgolovin: please do not suggest tests like this of methods that have no return statements, it always passes without errors and adds a little value to tests.

test/unit/pages/confirm/controller-test.js
```diff
@@ -1,31 +1,93 @@
...
+    });
+  });
+
+  describe('isConfigurationValid', function() {
+    it('should able to check Configuration Valid', function() {
```
dgolovin: You mean `should BE able to check configuration IS valid` to be correct, but it still adds nothing to understanding what is checked. It should sound like specification:
* should return true if nothin is detected and all components have been selected for installation;
* should return true if all detected components have supported versions;
* should return false if one of detected components have unsupported version;
* you name it and get coverage as high as you can! :)

dgolovin:
released with https://github.com/redhat-developer-tooling/developer-platform-install/pull/511.

PR #507, opened 2017-01-18
Progress bar refactoring

dgolovin:
* Pass whole ProgressStatus instance to the directive
* Use ng-repeat to show progress bars for components being installed
* Simplify call for angular digest when status or progress changes
* update ui and unit tests to pass


dgolovin:
test this

PR #509, opened 2017-01-19
JBDS-3869 DevSuite Installer should check/warn if Hyper-V is already enabled (win32)

sudhirverma:
JBDS-3869 DevSuite Installer should check/warn if Hyper-V is already enabled (win32)


sudhirverma:
test this

PR #514, opened 2017-01-23
CDK install never finishes installation

dgolovin:
Fix resolves cdk install issue and resolves eslint violations


dgolovin:
test this

PR #515, opened 2017-01-24
JBDS-4253 Show warning for virtualbox component if virtualization is not enabled or cannot be detected

sudhirverma:
Show warning for virtualbox component if virtualization is not enabled or cannot be detected

### DIFF
--- a/browser/pages/confirm/confirm.html
+++ b/browser/pages/confirm/confirm.html
@@ -355,6 +355,16 @@
       <span>Red Hat Container Development Kit requires Oracle VirtualBox {{checkboxModel.virtualbox.minimumVersion}} or newer 5.0.x release. Please uninstall detected Oracle VirtualBox and restart Installer to continue.</span>
     </div>
 
+    <div class="message-container error-message" ng-show="virtualization === false">
+      <span class="pficon pficon-error-circle-o"></span>
+      <span>Please enable hardware virtualization for your platform to improve performance.</span>
+    </div>
+
+    <div class="message-container error-message" ng-show="virtualization === undefined">
+      <span class="pficon pficon-error-circle-o"></span>
+      <span>Make sure hardware virtualization is enabled on your platform if supported to improve performance.</span>
+    </div>
+
     <!-- Cygwin -->
     <div id="cygwin-panel" class="panel panel-default panel2pxborder" ng-show="platform==='win32'" ng-class="{'zero-border': checkboxModel.cygwin.hasOption('detected') && checkboxModel.cygwin.selectedOption === 'detected'}">
       <div id="cygwin-panel-heading" class="panel-heading panel-normal"
--- a/browser/pages/confirm/controller.js
+++ b/browser/pages/confirm/controller.js
@@ -20,6 +20,7 @@ class ConfirmController {
     $scope.checkboxModel = {};
     $scope.platform = Platform.OS;
     $scope.detectionStyle = false;
+    $scope.virtualization = true;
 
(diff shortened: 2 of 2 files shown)


dgolovin:
test this

sudhirverma:
test this

browser/pages/confirm/controller.js
```diff
@@ -20,6 +20,11 @@ class ConfirmController {
     $scope.checkboxModel = {};
     $scope.platform = Platform.OS;
     $scope.detectionStyle = false;
+    $scope.virtualization = true;
+    Platform.isVirtualizationEnabled().then(status => {
```
dgolovin: It looks good, but should be running along with other detectors at line 74 down below.

dgolovin:
test this

browser/pages/confirm/controller.js
```diff
@@ -71,6 +72,10 @@ class ConfirmController {
       ).catch(
         ()=>this.setIsDisabled()
       );
+      Platform.isVirtualizationEnabled().then(status => {
```
dgolovin: Almost there. Now detection for installed components and for virtualization would run at the same time, but update on angular ui would happen twice. One in this.setIsDisabled() and second one in lambda after virtualization detector is finished. Lets move it to line 62 and add to detectors array first.
``` JavaScript
detectors.push(Platform.isVirtualizationEnabled().then(status => {
 $scope.virtualization = status;
});
```

$scope.apply() call is not required, because it is going to be done inside `this.setIsEnabled()`.

sudhirverma:
test this

sudhirverma:
test this

browser/pages/confirm/confirm.html
```diff
@@ -355,6 +355,16 @@
...
     </div>
 
+    <div class="message-container error-message" ng-show="virtualization === false">
+      <span class="pficon pficon-error-circle-o"></span>
+      <span>Please enable hardware virtualization for your platform to improve performance.</span>
+    </div>
+
+    <div class="message-container error-message" ng-show="virtualization === undefined">
+      <span class="pficon pficon-error-circle-o"></span>
+      <span>Make sure hardware virtualization is enabled on your platform if supported to improve performance.</span>
```
jrichter1: I would argue that even though any sort of performance is better than nothing, the error messages are sort of missing the point. Which is - virtual machines just won't run without hardware virtualization enabled. It isn't really a matter of performance. Can you get these messages to indicate that the user needs CPU virtualization enabled for the hypervisor to work properly?
(btw. sorry for neglecting this PR for so long)

dgolovin: I have not pushed it yet, because I have to test it first, This PR is based on  https://www.virtualbox.org/manual/ch01.html#features-overview stating:

>No hardware virtualization required. For many scenarios, VirtualBox does not require the processor features built into newer hardware like Intel VT-x or AMD-V. As opposed to many other virtualization solutions, you can therefore use VirtualBox even on older hardware where these features are not present

We could change the message to point out it might not work in some cases.

dgolovin: Verified using Windows 10 Pro host with disabled Virtualization in BIOS. It works as advertised but there are several cons:
1. Only 32-bit VM can be created
2. minishift fails to start with error: "This computer doesn't have VT-X/AMD-v enabled. Enabling it in the BIOS is mandatory".

dgolovin:
Rewired this PR according resent findings and sent new one https://github.com/redhat-developer-tooling/developer-platform-install/pull/595

PR #532, opened 2017-01-27
JBDS-3951 Installer needs to add the VirtualBox install directory to the PATH

sudhirverma:
Installer needs to add the VirtualBox install directory to the PATH

### DIFF
--- a/browser/model/virtualbox.js
+++ b/browser/model/virtualbox.js
@@ -131,6 +131,7 @@ class VirtualBoxInstallWindows extends VirtualBoxInstall {
           return this.configure(installer);
         })
         .then((result) => {
+          Platform.addToUserPath([this.option['install'].location]);
           return installer.succeed(result);
         })
         .catch((error) => {
--- a/test/unit/model/virtualbox-test.js
+++ b/test/unit/model/virtualbox-test.js
@@ -29,6 +29,7 @@ describe('Virtualbox installer', function() {
   let version = '5.1.12';
   let revision = '112440';
   let finalUrl = 'http://download.virtualbox.org/virtualbox/5.1.12/VirtualBox-5.1.12-112440-Win.exe';
+  let item2;
 
   installerDataSvc = sinon.stub(new InstallerDataService());
   installerDataSvc.getRequirementByName.restore();
@@ -45,6 +46,7 @@ describe('Virtualbox installer', function() {
     infoStub = sinon.stub(Logger, 'info');
     errorStub = sinon.stub(Logger, 'error');
     sha256Stub = sinon.stub(Hash.prototype, 'SHA256', function(file, cb) { cb('hash'); });
+    item2 = new InstallableItem('jdk', 1000, 'url', 'installFile', 'targetFolderName', installerDataSvc);
 
     mockfs({
       tempDirectory: {},
@@ -133,8 +135,9 @@ describe('Virtualbox installer', function() {
     let downloadedFile = path.join('tempDirectory', 'virtualbox.exe');
 
     describe('on windows', function() {
-      let installer;
+      let installer, helper;
       beforeEach(function() {
+        helper = new Installer('virtualbox', fakeProgress);
         sandbox.stub(Platform, 'getOS').returns('win32');
         installer = new VirtualBoxInstallWindows(version, revision, installerDataSvc, downloadUrl, 'virtualbox.exe', 'virtualbox', 'sha');
         installer.ipcRenderer = { on: function() {} };
@@ -161,7 +164,6 @@ describe('Virtualbox installer', function() {
       });
 
       it('setup should wait for all downloads to complete', function() {
-        let helper = new Installer('virtualbox', fakeProgress);
         let spy = sandbox.spy(installer, 'installMsi');
 
         installerDataSvc.downloading = true;
@@ -174,7 +176,6 @@ describe('Virtualbox installer', function() {
 
       describe('configure', function() {
         it('should call installMsi if all downloads have finished', function() {
-          let helper = new Installer('virtualbox', fakeProgress);
           let spy = sandbox.spy(installer, 'installMsi');
           sandbox.stub(child_process, 'execFile').yields();
 
@@ -228,7 +229,6 @@ describe('Virtualbox installer', function() {
     it('should catch errors during the installation', function(done) {
       sandbox.stub(child_process, 'execFile').yields(new Error('critical error'));
       sandbox.stub(child_process, 'exec').yields(new Error('critical error'));
-      let item2 = new InstallableItem('jdk', 1000, 'url', 'installFile', 'targetFolderName', installerDataSvc);
       item2.setInstallComplete();
       item2.thenInstall(installer);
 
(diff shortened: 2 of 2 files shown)


sudhirverma:
test this

sudhirverma:
test this

test/unit/model/virtualbox-test.js
```diff
@@ -240,10 +241,17 @@ describe('Virtualbox installer', function() {
       }
     });
 
+    it('should able to check file were extracted to temp', function(done) {
+      sandbox.stub(child_process, 'execFile').yields(null, "Files were extracted to temp", "");
+      item2.setInstallComplete();
+      item2.thenInstall(installer);
+      installer.install(fakeProgress, success, failure);
```
dgolovin: call installer.installAfterRequirements() and you would not need to create jdk installer at all.

test/unit/model/virtualbox-test.js
```diff
@@ -240,10 +241,17 @@ describe('Virtualbox installer', function() {
...
+      done();
```
dgolovin: if you use done callback from mocha, you should be calling it from callback passed to installer.installAfterRequirements().

``` JavaScript
installer.installAfterRequirements(fakeProgress, function (){
 expect().something.has.been.done;
 done();
 }, function(){
  expect.fail();
 });
```
that is needed only if you really expect async code is still going to be running when test method exits.

if you mock all async methods then you don't need use done callback at all, just `expect().test.the.condition;` should do the trick.

test/unit/model/virtualbox-test.js
```diff
@@ -240,10 +241,17 @@ describe('Virtualbox installer', function() {
       }
     });
 
+    it('should able to check file were extracted to temp', function(done) {
```
dgolovin: Did you mean to name it as 'should run virtualbox installer executable to extract msi installer into temp folder'?

browser/model/virtualbox.js
```diff
@@ -131,6 +132,9 @@ class VirtualBoxInstallWindows extends VirtualBoxInstall {
           return this.configure(installer);
         })
         .then((result) => {
+          if(result && this.targetDir) {
```
dgolovin: just replace to 
``` JavaScript
Platform.addToUserPath(this.getOption('install').location);
```

browser/model/virtualbox.js
```diff
@@ -173,10 +177,10 @@ class VirtualBoxInstallWindows extends VirtualBoxInstall {
       // msiexec logs are in UCS-2
       Util.findText(path.join(this.installerDataSvc.installDir(), 'vbox.log'), 'CopyDir: DestDir=', 'ucs2').then((result)=>{
         let regexTargetDir = /CopyDir: DestDir=(.*)\,.*/;
```
dgolovin: these changes don't needed here

browser/model/virtualbox.js
```diff
@@ -16,6 +16,7 @@ class VirtualBoxInstall extends InstallableItem {
     this.minimumVersion = version;
     this.maximumVersion = '5.2.0';
     this.revision = revision;
+    this.targetDir = "";
```
dgolovin: You don't need new class field, see my comments below. Just use location property from 'install' option as it is explained below.

test/unit/model/virtualbox-test.js
sudhirverma: yes

browser/model/virtualbox.js
sudhirverma: Can I use this.getOption('install').location because getOption is not even declare any where.

dgolovin:
``` JavaScript
this.option('install').location
```

sudhirverma:
test this

sudhirverma:
test this

browser/services/platform.js
```diff
@@ -136,7 +136,7 @@ class Platform {
   static addToUserPath_win32(locations) {
     return Platform.getUserPath_win32().then((pathString)=>{
       let pathLocations = pathString.split(';');
-      return Platform.setUserPath_win32([...locations.filter(item=>!pathLocations.includes(item)), pathString].join(';'));
+      return Platform.setUserPath_win32([...[locations].filter(item=>!pathLocations.includes(item)), pathString].join(';'));
```
dgolovin: locations parameter is an array, lets keep it that way. I want to be able to add several locations in single call.

browser/model/virtualbox.js
```diff
@@ -131,6 +131,9 @@ class VirtualBoxInstallWindows extends VirtualBoxInstall {
           return this.configure(installer);
         })
         .then((result) => {
+          if(result) {
```
dgolovin: This check is not required, installation ether succeeded and then callback is executed or failed and catch callback is executed.

browser/model/virtualbox.js
```diff
@@ -131,6 +131,9 @@ class VirtualBoxInstallWindows extends VirtualBoxInstall {
...
+            Platform.addToUserPath(this.option['install'].location);
```
dgolovin: argument must be an array
``` JavaScript
Platform.addToUserPath([this.option['install'].location]);
```

test/unit/model/virtualbox-test.js
```diff
@@ -240,10 +240,23 @@ describe('Virtualbox installer', function() {
       }
     });
 
+    it('should run virtualbox installer executable to extract msi installer into temp folder', function(done) {
+      let stub = sandbox.stub(child_process, 'execFile').yields(null, "Files were extracted to temp", "");
+      try {
```
dgolovin: Try/catch is not required here, please remove it.

test/unit/model/virtualbox-test.js
```diff
@@ -240,10 +240,23 @@ describe('Virtualbox installer', function() {
...
+        installer.install(fakeProgress, success, failure);
+        done();
```
dgolovin: calling done inside main thread is wrong, it should be inside callback, please remove it here and form it second parameter callback.

test/unit/model/virtualbox-test.js
```diff
@@ -240,10 +240,23 @@ describe('Virtualbox installer', function() {
...
+      } catch (error) {
+        expect.fail();
+      }
+      return helper.execFile().then(() => {
```
dgolovin: if you stub everything properly all the work should be done inside main thread an you can just check with expect() here, there is no need to return promise instance from it.

test/unit/model/virtualbox-test.js
```diff
@@ -240,10 +240,23 @@ describe('Virtualbox installer', function() {
...
+        expect(stub).to.be.calledOnce;
+        expect(stub).to.be.true;
```
dgolovin: this looks wrong, you instead should be checking that stubbed method called with right parameters. In your case you should check that downloaded executable file is called to unpack to temp folder.

sudhirverma:
test this

sudhirverma:
test this

dgolovin:
Included into https://github.com/redhat-developer-tooling/developer-platform-install/pull/544.

PR #546, opened 2017-02-08
Increase unit tests code coverage confirm controller

sudhirverma:


test/unit/pages/confirm/controller-test.js
```diff
@@ -104,6 +105,29 @@ describe('ConfirmController', function() {
     });
   });
 
+  describe('isConfigurationValid', function() {
```
dgolovin: isConfigurationValid() enables 'Next' button on confirmation page. Current implementation is a condition that takes into consideration only products with dependencies. There are only two of them:
* DevStudio;
* CDK;

everything else is those two products requirements.

DevStudio considered configured correctly if it is selected for installations and all dependencies configured correctly or it is not selected for installation. The same statement is valid about CDK. Dependency is configured correctly if it is selected for installation or it is detected with valid version.

DevStudio only dependency is:
* OpenJDK or any java (valid if semver.satisfies('>=1.8.0','<1.9.0') );

CDK dependencies are:
* virtualbox (valid if semver.satisfies('>=5.1.26','<5.2.0') );
* cygwin (valid if openssh and rsync packages installed);

For isConfigurationValid() function I would suggest this specification:
* should return true if all components selected for install are configured correctly
* should return false if nothing is selected for installation 
* should return false if at least one component selected for installation is not configured correctly

Implementation can be done as white box test meaning you are using direct stubbing for involved functions. Another approach is to play with real installers and move them into one of three states:
* selected for install;
* skipped or detected and has valid version;
* detected and has invalid version;

then test if isConfigurationValid() returns expected value. You can use addOption() method and selectedOption property to do that. For example if you stub detectExistingInstall() methods for all installers to suppress detection and then call for every installer
``` JavaScript
installer.addOption('install', undefined, 'location', true); 
installer.selectedOption = 'install';
```
you would select all components for installation and isConfigurationValid() would expect to return true.

If you call 
``` JavaScript
installer.selectedOption = 'detected';
```
for every installer you would deselect everything and isConfigurationValid() would expect to return false.

Then you have to figure out how to implement third it() method. :)

test/unit/pages/confirm/controller-test.js
```diff
@@ -104,6 +105,29 @@ describe('ConfirmController', function() {
...
+    it('should able to check JBDS is configured', function() {
+      let stub = sandbox.stub(ConfirmController.prototype, 'jbdsIsConfigured').returns(true);
```
dgolovin: First you stub it to return true

test/unit/pages/confirm/controller-test.js
```diff
@@ -104,6 +105,29 @@ describe('ConfirmController', function() {
...
+      confirmController.isConfigurationValid();
+      expect(stub).calledOnce;
+      expect(confirmController.jbdsIsConfigured()).to.be.true;
```
dgolovin: Then you check if stabbed method returns true. What are you testing? It looks like you testing if sinon stubbing is working correctly :) Instead you should stub all three involved methods and that would be white box testing.  Se my comment above and try to implement test with second approach.

PR #562, opened 2017-02-17
JBDS-4066 Improve error reporting on account info page

dgolovin:
Fix:
- moves validation logic to controller;
- adds reset for error reported after submit if any field is edited;
- elements location within page, so nothing moves after error appears.


jrichter1:
+1

PR #568, opened 2017-02-21
JBDS-4066 DevSuite Installer Account step styles are inconsistent with Target Folder step

sudhirverma:


jrichter1:
login

jrichter1:
Could you try to align the entire form so the input fields have a proper length?

PR #569, opened 2017-02-21
JBDS-4066 DevSuite Installer Account step styles are inconsistent

dgolovin:


dgolovin:
Login page error under windows

login-err-win

PR #579, opened 2017-02-27
JBDS-3825 Cygwin installs into a different folder when some non-ascii  characters are present in folder name

sudhirverma:
characters are present in folder name

### DIFF
--- a/browser/directives/pathValidator.js
+++ b/browser/directives/pathValidator.js
@@ -5,6 +5,7 @@ import path from 'path';
 import Platform from '../services/platform';
 
 let pathWindowsRegex = /^[a-zA-Z]:\\(?:[^\\/:*?"<>|\r\n]+\\)*[^\\/:*?"<>|\r\n]*$/;
+let nonAsciiRegex = /[^\x00-\x7F]+/;
 
 function pathValidator() {
   return {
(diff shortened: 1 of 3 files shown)


browser/directives/pathValidator.js
```diff
@@ -42,6 +47,7 @@ function pathValidator() {
       mCtrl.$validators['notAbsolute'] = isAbsolute;
       mCtrl.$validators['tooLong'] = validateLength;
       mCtrl.$validators['hasSpaces'] = hasNoSpaces;
+      mCtrl.$validators['invalidAscii'] = validateAscii;
```
dgolovin: It affects windows only, so it is really should be inside if below.

browser/directives/pathValidator.js
```diff
@@ -13,10 +14,14 @@ function pathValidator() {
 
       function validateFormatWindows(value) {
         let trimmedValue = value.trim();
-
         return pathWindowsRegex.test(trimmedValue);
       }
 
+      function validateAscii(value) {
+        let trimmedValue = value.trim();
```
dgolovin: Trimming is not required, because angular automatically trims text fields, unless explicitly said 'do not trim' through ngTrim directive. See details here https://docs.angularjs.org/api/ng/input/input%5Btext%5D.

browser/directives/pathValidator.js
```diff
@@ -13,10 +14,13 @@ function pathValidator() {
 
       function validateFormatWindows(value) {
         let trimmedValue = value.trim();
```
dgolovin: This line is not needed as well

PR #584, opened 2017-03-02
JBDS-3828 Virtualbox fails to install when user has non-ascii characters in the name

sudhirverma:


jrichter1:
I don't think just showing an error and preventing the installation should be the way to go in this case. The thing is, the installation fails due to the files being extracted into temp folder which is in the user home containing non-ascii chars. But we are the ones who put those files there.

So my proposal is - use the detection you've implemented and if the username contains such characters, just extract the vbox installer somewhere safe, run it and then get rid of it. Don't just show an error, because this is something we can in fact fix ourselves :)

dgolovin:
I agree and we already discussed it with @sudhirverma, but have not decided yet what way to go. Probably good option is to use targetFolder/temp and then remove it, it should be non-ascii chars free anyway or cygwin and devstudio would fail installation.

dgolovin:
Just tested it under user 'Śomeone'  with overridden  TMP with new location w/o non-ascii characters. Installation worked without any problems. CDK 3 works fine even after I restored TMP to its original value with c:\Users\Śomeone\AppData\Temp.

dgolovin:
Update: msiexec does not pick up env variables passed as options.env to child_process.exe or child_process.execFile. The same behavior is observed when running 'set TMP=path/to/ne/temp' before running msiexec in the same console. Looks like msiexec uses TMP variable defined in Control Panel only.

These all means Installer cannot workaround it by setting TMP env variable.

Validation should actually check the value for %TMP% not %CURRENTPROFILE% because user can override it in control panel and installation for Virtualbox will work.

browser/pages/location/controller.js
```diff
@@ -40,6 +42,23 @@ class LocationController {
     this.checkFolder();
   }
 
+  checkUserProfilestatus() {
+    return new Promise((resolve, reject)=> {
+      if (Platform.OS === 'win32') {
+        Util.executeCommand('echo %USERPROFILE%',1).then((path)=>{
```
dgolovin: should check for %TMP% value instead in case it is set directly in control panel.

browser/pages/location/location.html
```diff
@@ -23,21 +23,29 @@
...
               <span class="pficon pficon-info"></span>
               The selected folder will be created.
             </strong>
           </div>
 
-          <div ng-show="locationForm.folder.$error" class="help-block has-error">
+          <div ng-show="locCtrl.checkUserProfilestatus() && locCtrl.nonasciichars" class="help-block has-error">
+            <strong id="checkUserProfilestatus" class="help-block has-error">
+              <span class="pficon-error-circle-o"></span>
+              Path to user home directory has Non-ASCII characters
```
dgolovin: Message should say: 'TMP environment variable points to location with Non_ASCII characters in name, it could lead to errors during installation, please change it to location that uses only ASCII characters'.

jrichter1: Wouldn't it be easier to just change the folder here 
https://github.com/redhat-developer-tooling/developer-platform-install/blob/master/browser/model/virtualbox.js#L121
to something we know is safe if there is an unsafe character in it?

dgolovin: This is not an issue :) This code extracts MSI file out of EXE file. It works without any issues. Virtualbox MSI installer is executed with msiexec it has problems when extracting CAB files from MSI into %TMP% location. I tried to use options.env.TMP in execFile to override it, but msiexec just ignors it. The same happens when I do it in cmd.exe terminal. I tried to use 'set TMP=c:\Temp'  , but msi exec anyway uses %TMP% value set in system settings.

jrichter1: The issue here is we're actively extracting the MSI file + CAB files into %TMP%. All I'm saying here is, we can extract it somewhere else if %TMP% has unsafe characters in it to avoid the CAB file not found problem.

dgolovin: You can extract MSI in different location, but when you run msi as 

``` shell 
msiexec /i installer.msi
```

there is no good way to control where msiexec extracts .cab files during installation. They go to %TMP% and code below does not work, because MSI seems to read %TMP% directly form configuration not from current session. For example code below does not work.

``` shell
set TMP=c:\Temp
msiexec /i installer.msi
```

msi installer itself does not have public property to override TMP folder location.

We can only set TMP env var directly in settings using powershell, but that would require to restore original value after installation is over. It is also not easy to do, because original value normally %USERPROFILE%/AppData/Local/Temp and to save that string without %VAR% expansion it should be read directly form registry, so it is a little to much :)

I am out of ideas and open for any working solution.

dgolovin: Another alternative is to move this verification to confirmation page and do validation there and show error message only if Virtualbox is going to be installed. The message can suggest to user manually set TMP in system preferences and restart installer. That validation would work only in case of virtual box installation.

jrichter1: I'm not following what your problem here is so I tried it out like this:
 - logged in as Pakoň (the original troublemaker)
 - downloaded vbox exe
 - opened cmd and did this:

$ virtualbox_5.1.12.exe --extract -path c:\temp
 - this extracted the msi and common.cab to c:\temp

$ msiexec /i c:\temp\VirtualBox-5.1.12-r112440-MultiArch_amd64.msi /qn INSTALLDIR="c:\temp\vbox" /norestart
 - now I have a working virtualbox

Anything I'm missing here?

dgolovin: Nope, I am the one who is missing ;) Basically you suggest to do this https://github.com/redhat-developer-tooling/developer-platform-install/pull/592. Just verified it and it works for me.
👍

jrichter1: Yes, that's the idea.

dgolovin:
replaced with https://github.com/redhat-developer-tooling/developer-platform-install/pull/592

PR #586, opened 2017-03-03
Fix unit test errors on macOs

dgolovin:


test/unit/model/installable-item-test.js
```diff
@@ -232,11 +232,25 @@ describe('InstallableItem', function() {
...
 
   describe('setOptionLocation method', function() {
-    it('should set location property for specific option', function() {
+    it('should set location property for specific option if it is present', function() {
       item.addOption('detected');
       item.setOptionLocation('detected', 'location');
       expect(item.option.detected.location).to.be.equal('location');
     });
+
+    it('should set location property for specific option if it is present', function() {
```
jrichter1: I guess the two test cases should not really have the same title, can you fix that?

dgolovin: You are correct, that is supposed to test if branch for case when option is not present. I'll fix it, thanks!

PR #587, opened 2017-03-06
Rename jbds to devstudio

jrichter1:
Since JBDS is forbidden for some time and sometimes we called it devstudio anyway. Initially only wanted to change element IDs but since some of them are loaded from requirements.json, I just replaced all jbds with devstudio.


dgolovin:
:+1:

dgolovin:
Rebased on top of the master and included into new PR https://github.com/redhat-developer-tooling/developer-platform-install/pull/588.

PR #589, opened 2017-03-07
Fix e2e test error on page transition

jrichter1:


test/system/test-base.js
```diff
@@ -222,6 +222,15 @@ function systemTest() {
                         fail(product + 'failed to install');
                         done();
                       }
+                    }).catch(function(error) {
+                      if (error.message.indexOf('element is not attached') > -1) {
+                        browser.getLocationAbsUrl().then(function(url) {
+                          if (url !== '/start') {
+                            throw error;
+                            done();
```
dgolovin: This done() call seems never going to happen, because of throw before it.

jrichter1: right

test/system/test-base.js
```diff
@@ -222,6 +222,14 @@ function systemTest() {
                         fail(product + 'failed to install');
                         done();
                       }
+                    }).catch(function(error) {
+                      if (error.message.indexOf('element is not attached') > -1) {
+                        browser.getLocationAbsUrl().then(function(url) {
+                          if (url !== '/start') {
+                            throw error;
+                          }
```
dgolovin: Should  it call done() in case url == '/start'?

jrichter1: Not really, it will just skip into the next iteration and the /start handling is right at the start of that.

dgolovin: LGTM 👍

PR #592, opened 2017-03-08
JBDS-3828 Virtualbox fails to install

dgolovin:


dgolovin:
common.cab and *.msi files should be removed after installation.

PR #610, opened 2017-03-16
JBDS-4313 Show installer window when content is loaded and ready

dgolovin:


dgolovin:
Here are the demos how it worked before and after fix
[IBefore fix](https://youtu.be/Oi2dWGlJq9k) [After fix](https://youtu.be/tTT-6KRe3MM)

PR #620, opened 2017-03-20
JBDS-4301 Allow loading subset of components from requirements.json

jrichter1:
Rewritten the hardcoded init for model classes into something more "dynamic". Quite a big change, so here is how it works:
 - everything (almost) is now loaded from requirements.json - it maps strings to class names and instantiates them using javascript magic (see ComponentLoader and DynamicClass for reference)
 - there is some default ordering of the installation sequence if all the components are present
 - if some component is omitted, the loader squashes the sequence = replaces the component with the components that depended on it (e.g. if cygwin is omitted, cdk installs after virtualbox/hyper-v), the same principle is used for the checkboxModel in confirmController
 - the model constructors are fed arguments directly from requirements.json, therefore the ordering of the properties in requirements.json now matters
   - if a new argument to a constructor is needed, it should be appended to the end of the argument list
 - confirm page now only shows components that are instantiated
 - moved some requirements exclusively to their respective platforms (like cygwin)
 - one of virtualbox/hyper-v is now removed from components upon clicking "install" depending on what option is available. I tried to do this sooner, but then they would need to be re-added if the user changed the settings and detection was run again

### DIFF
--- a/browser/services/componentLoader.js
+++ b/browser/services/componentLoader.js
@@ -1,0 +1,97 @@
+'use strict';
+
+import InstallerDataService from './data';
+
+const baseOrder = {
+  'root': ['jdk'],
+  'jdk':  ['virtualbox', 'devstudio'],
+  'virtualbox': ['hyperv'],
+  'hyperv': ['cygwin'],
+  'cygwin': ['cdk'],
+  'devstudio': [],
+  'cdk': []
+};
+
+class ComponentLoader {
+  constructor(installerDataSvc) {
+    this.requirements = installerDataSvc.requirements;
+    this.installerDataSvc = installerDataSvc;
+  }
+
+  loadComponents() {
+    for (let key in this.requirements) {
+      this.addComponent(key);
+    }
+    this.orderInstallation();
+  }
+
+  loadComponent(key) {
+    this.addComponent(key);
+    this.orderInstallation();
+  }
+
+  removeComponent(key) {
+    this.installerDataSvc.allInstallables().delete(key);
+    this.orderInstallation();
+  }
+
+  addComponent(key) {
+    if (this.requirements[key] && this.requirements[key].bundle !== 'tools') {
+      let skippedProperties = ['name', 'description', 'bundle', 'vendor', 'virusTotalReport', 'modulePath'];
+      let args = [this.installerDataSvc];
+      if (this.requirements[key].dmUrl) {
+        skippedProperties.push('url');
+      }
+
+      for (let property in this.requirements[key]) {
+        if (skippedProperties.indexOf(property) < 0) {
+          args.push(this.requirements[key][property]);
+        }
+      }
+      let component = new DynamicClass(this.requirements[key].modulePath, args);
+
+      this.installerDataSvc.addItemToInstall(key, component);
+    }
+  }
+
+  orderInstallation() {
+    let newOrder = {};
+    Object.assign(newOrder, baseOrder);
+    let changed;
+
+    do {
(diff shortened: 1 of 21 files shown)


jrichter1:
And one more side effect is that progress bars are now ordered the same way keys are ordered in requirements.json.

jrichter1:
@dgolovin @sudhirverma can you guys properly break this (I mean test this) if you find the time? Comments, complaints, corrections welcome as always. Thanks.

browser/services/componentLoader.js
```diff
@@ -0,0 +1,126 @@
...
+        let nextItem = this.installerDataSvc.getInstallable(newOrder[key][i]);
+        value.thenInstall(nextItem);
+      }
+    }
+  }
+}
+
+class DynamicClass {
+  constructor (className, opts) {
+    return new classes[className](...opts);
```
dgolovin: why not something like (pardon me if I am putting something not working here, I have not tried it yet)

``` JavaScript 
new require(jsonData.modulePath)(jsonData);
```
considering that we have object properties defined inside requirements.json

dgolovin:
Would be also good to use something like https://www.npmjs.com/package/toposort to flatten requirement instead of doing it ourselves.

browser/services/componentLoader.js
```diff
@@ -0,0 +1,126 @@
...
+    this.orderInstallation();
+  }
+
+  addComponent(key) {
+    if (this.requirements[key] && this.requirements[key].bundle !== 'tools') {
+      let skippedProperties = ['name', 'description', 'bundle', 'vendor', 'virusTotalReport'];
+      let folder = key;
+      switch (key) {
+        case 'jdk':
+          folder = 'jdk8'
```
jrichter1: True, and makes feel dumb for not doing it :)

browser/services/componentLoader.js
jrichter1: Will have to try this one out

jrichter1: This can only be done if the classes define 
```javascript
module.exports = Class
```
and we already have different export mechanism in place so we can use imports. The question is whether it would be beneficial to rewrite exports to module.exports for this.

jrichter1:
Ad flattening: With two parallel branches, flattening the graph would degrade the structure too much - it would still need some algorithm to flatten the branches individually.
This is more like bypassing inactive nodes and I couldn't find a nice library for that (yet).

browser/services/componentLoader.js
jrichter1: I actually found something like this can be done:
```javascript
let klass = require(`../${modulePath}`);
return new klass.default(...opts);
```

jrichter1: That is if we keep using default exports.

dgolovin:
You right if we are going to keep parallel installations, but if we don't that should work. I know that performance would be degrading if we install everything one by one, mostly because devstudio install is really slow and it is good to have it running in parallel with everything.

jrichter1:
Well, if we were to install everything one by one, there would be no need to bother with this graph structure in the first place. A properly ordered array would do the job just fine.
The question is, what is the preferred solution.

dgolovin:
There are several reasons to get rid of it:
1. MSI does not support parallel installation
2. It takes page space to visualize all installations in progress
3. we have to limit number of simultaneous install anyway as nuber of components available for installation grow
4. (1), (2), (3) complicates dependency resolution and install because to many constrains

But lets keep it out of this issue. We can address it later.

PR #628, opened 2017-03-22
JBDS-4314 Hyper-v shows up on Install page on macOS

dgolovin:
Fix set selectedOption as detected, which works as skip swithc
when there is no detected option in option collection.


jrichter1:
Wouldn't this cause another component to be falsely counted as detected? Also would something like this work instead #627 ?

PR #645, opened 2017-03-29
JBDS-4303 Show warning for virtualbox component if virtualization is not enabled or cannot be detected (macos)

sudhirverma:


browser/pages/confirm/confirm.html
```diff
@@ -374,11 +374,16 @@
       <span>Red Hat Container Development Kit requires Oracle VirtualBox {{checkboxModel.virtualbox.minimumVersion}} or newer 5.1.x release. Please uninstall detected Oracle VirtualBox and restart Installer to continue.</span>
     </div>
 
-    <div class="message-container error-message" ng-show="!confCtrl.isDisabled && !checkboxModel.hyperv.isConfigured() && checkboxModel.virtualbox.virtualizationEnabled === false && (checkboxModel.cdk.selectedOption == 'install' || checkboxModel.virtualbox.selectedOption == 'install')">
+    <div class="message-container error-message" ng-show="!confCtrl.isDisabled && !checkboxModel.hyperv.isConfigured() && checkboxModel.virtualbox.virtualizationEnabled === false && (checkboxModel.cdk.selectedOption == 'install' || checkboxModel.virtualbox.selectedOption == 'install') && platform === 'win32'">
```
dgolovin: Check for platform should not be required, because we still following the same rules:
1. Promise resolved to false - error;
2. Promise resolved to true - everything is fine;
3. Promise resolved to undefined - warning.

Even though we going to have warning on macOS when virtualization is supported, it is fine, because we cannot say it is turned on or not.

sudhirverma: Yes I know that. I have used platform because I want to show different error message in windows and mac if virtualization not found.

dgolovin: You right it does not make sense to mention BIOS for macOS.

PR #678, opened 2017-04-26
JBDS-4372 Remove dots from minishift filename

jrichter1:
The extension check fails on mac with dots in file name.


jrichter1:
test this

PR #685, opened 2017-05-03
JBDS-4328 Move 'Detection is in progress' indicator in the middle of the screen

sudhirverma:


jrichter1:
It's in a nice spot, but can you make it even more visible? Like even 3 times as large and with a colour that is not as grey-ish as everything in the background. Not sure what choice we have on colours, maybe light blue? WDYT?

jrichter1:
And you could even try to move the detection message along with the spinner. Just an idea to consider, it might look better :)

dgolovin:
See linked PR for details

PR #687, opened 2017-05-05
JBDS-4375 Show remaining time for downloads

jrichter1:
Shows an ETA for each download.


jrichter1:
I guess that depends on preference, but sure, we can have a little Firefox vs Chrome format battle :)

jrichter1:
@dgolovin I guess it does look better after all, feel free to try it out now

PR #696, opened 2017-05-11
JBDS-4387 Add current user to Hyper-V Administrators Local Group

dgolovin:


jrichter1:
Works fine, requires the user to log out and back in to refresh the group though. I can't find any way to do that without relog/reboot so I guess we need to document this. Right, @dgolovin ?

dgolovin:
Sounds right @jrichter1, what we can also do is to add info message under CDK 3 item. That would say: "You need to logoff/login after installation to run CDK without Administrative privileges".

dgolovin:
@jrichter1, but lets keep it in separate issue.

PR #697, opened 2017-05-12
Remove Alpha qualifier from mac

jrichter1:
Going GA we are


dgolovin:
👍 Thanks @jrichter1 !

PR #704, opened 2017-05-17
Fixed kompose failure on mac

sudhirverma:


browser/model/kompose.js
```diff
@@ -24,7 +24,7 @@ class KomposeInstall extends InstallableItem {
     let installer = new Installer(KomposeInstall.KEY, progress, success, failure);
     let komposeExe = path.join(komposeDir, Platform.OS === 'win32' ? 'kompose.exe' : 'kompose');
     return Promise.resolve().then(()=> {
-      if(this.downloadedFile.endsWith('.exe') || path.parse(this.downloadedFile).ext == '') {
+      if(this.downloadedFile.endsWith('.exe') || this.downloadedFile.endsWith('')) {
```
jrichter1: Every defined string ends with an empty string, including an empty string itself. In other words, this check is useless. For a quick fix you might just replace dots in kompose filename for underscores for example.

PR #714, opened 2017-05-25
JBDS-3632 Use prepackaged version for cygwing for bundle installer

dgolovin:


gulp-tasks/dist-win32.js
```diff
@@ -94,7 +95,18 @@ module.exports = function(gulp) {
 
   // prefetch cygwin to always include into installer
   gulp.task('prefetch-cygwin', ['create-prefetch-cache-dir'], function() {
-    return download.prefetch(reqs, 'always', config.prefetchFolder);
+    download.prefetch(reqs, 'always', config.prefetchFolder).then(function() {
+        return new Promise(function(resolve,reject) {
+          child_process.exec(`${path.resolve(config.prefetchFolder)}\\${reqs.cygwin.filename} -D --no-admin --quiet-mode --only-site -l ${path.resolve(config.prefetchFolder,'packages')} --site http://mirrors.xmission.com/cygwin --categories Base --packages openssh,rsync --root ${path.resolve(config.prefetchFolder,'packages')}`,function(error, std, err) {
```
jrichter1: Wouldn't it make more sense to move this to a stage that is not part of package-simple task? It is pointless when building just the online installer. Just something to consider...

PR #716, opened 2017-05-26
JBDS-4414 Add uninstaller into control panel

dgolovin:


jrichter1:
Looks good, with 2 suggestions:
 - use the full product name "Red Hat Development Suite" - we gotta stay official :)
 - maybe remove the waiting for input at the end of uninstall script - I guess we'll discourage people from using it directly now and it looks weird when calling it from the control panel

dgolovin:
Sounds good to me @jrichter1, I'll update PR, thanks!

dgolovin:
Updated PR to have appropriate name and added install location to comments attribute to at least be able to distinguish on install from another.

PR #727, opened 2017-06-05
Update devstudio link + remove duplicated build steps

jrichter1:
- Something/someone keeps deleting devstudio builds from the staging server.
 - update-requirements was called twice in most cases, update-package on mac is now part of update-requirements, so I deleted that one as well


jrichter1:
I've realized now that everything is pulled from content.json, we can just use the latest symlink

PR #732, opened 2017-06-07
JBDS-3716 Improve all in one installer performance

dgolovin:
Result is new bundled installer starts 3 times faster than old one, but it is 70Mb bigger.

### DIFF
--- a/gulp-tasks/config.js
+++ b/gulp-tasks/config.js
@@ -17,6 +17,7 @@ let config = {
 config.buildFolderRoot = path.join('dist', config.artifactPlatform + '-' + config.artifactArch );
 config.buildFileNamePrefix = config.artifactName;
 config.buildFolderPath = path.resolve(config.buildFolderRoot);
+config.genFolderPath = path.join(config.buildFolderPath, 'devsuite-win32-x64');
 config.configIcon = path.resolve(path.join('resources', config.artifactName + '.ico'));
 
 module.exports = config;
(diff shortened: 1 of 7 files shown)


gulp-tasks/config.js
```diff
@@ -17,6 +17,7 @@ let config = {
 config.buildFolderRoot = path.join('dist', config.artifactPlatform + '-' + config.artifactArch );
 config.buildFileNamePrefix = config.artifactName;
 config.buildFolderPath = path.resolve(config.buildFolderRoot);
+config.genFolderPath = path.join(config.buildFolderPath, 'devsuite-win32-x64');
```
jrichter1: This variable looks like it is not being used at all.

jrichter1:
The blazing speed at which it unpacks is awesome. And it is still roughly the same size as before (without compression??).

gulp-tasks/config.js
dgolovin: Right, I was trying to skip app generation every time I run the build and forget to clean up. 
Another reason for it not to be here is it is windows specific.
I'll remove it.
Thanks @jrichter1!

dgolovin:
I can make it even the same size, but that would require to pack every bundled binary and then unpack it first before install. I think we can pay 70Mb in size gain for not doing that :)

dgolovin:
test this

jrichter1:
test this

jrichter1:
test this

PR #734, opened 2017-06-07
Remove devstudio uninstall shortcut

jrichter1:
I was thinking we could remove the uninstall shortcut since some people were getting confused by it. @dgolovin thoughts?


dgolovin:
@jrichter1 sounds good to me 👍
It always appears first when I search for DevStudio in Windows menu and that is confusing.
Now when uninstall can be triggered from control panel, we can skip creating separate uninstaller shortcut for DevStudio.

PR #747, opened 2017-06-15
Clone requirements when calling loadMetadata

jrichter1:
so that it can be called repeatedly on the same set of requirements. Basically, loadMetadata would change the data stored in require('requirements.json') and since require caches the result, it would never load the original requirements again. 
Now it should not touch the original content and instead create a working copy.


jrichter1:
test this

PR #748, opened 2017-06-15
Java detection for windows is broken

dgolovin:
Fix adjusts detection for java installation on macOS so it
does not affect detection for windows


dgolovin:
test this

PR #751, opened 2017-06-16
JBDS-4423 Fuse Tools installer integrated

dgolovin:


browser/model/devstudio.js
```diff
@@ -9,14 +9,15 @@ import Logger from '../services/logger';
 import JdkInstall from './jdk-install';
 
 class DevstudioInstall extends InstallableItem {
-  constructor(installerDataSvc, targetFolderName, downloadUrl, fileName, devstudioSha256, additionalLocations, additionalIus) {
-    super(DevstudioInstall.KEY, downloadUrl, fileName, targetFolderName, installerDataSvc, true);
+  constructor(keyName, installerDataSvc, targetFolderName, downloadUrl, fileName, sha256sum, additionalLocations, additionalIus, useDownload) {
```
jrichter1: Any particular reason to use keyName in constructor here, while other model classes use a combination of static Class.KEY and this.keyName?

dgolovin: the reason is reusing it for fuse tools installation with additionalIus (IU - Installation Unit) configured in addition

jrichter1:
I've run into a couple of problems while testing this:
 - fuse can be selected without devstudio (particularly deadly in online installer or if you did not bother with installing java)
 - fuse ignores bundled devstudio jar and always looks for it in the download folder, which is disastrous in the offline installer for a change

jrichter1:
Or rather it might be caused by the fact that devstudio filename gets updated in requirements.json during the build and fuse doesn't.

dgolovin:
Thanks Jan, I am looking into both issues. First is related to the fact that we have no generic dependency resolution. I've added 'requires' attribute to every components where applicable, but fact that CDK should work with ether virtualbox or hyper-v holding. Hope I'll find solution and resolve this issue till my PTO starts.

Second seems to be simple bug to fix, because you already found the reason for it.

dgolovin:
test this

PR #756, opened 2017-06-18
Update Installer version to 1.4.1.GA

dgolovin:


requirements.json
```diff
@@ -8,27 +8,27 @@
     "platform": {
       "win32": {
         "bundle": "yes",
-        "dmUrl": "https://developers.redhat.com/download-manager/jdf/file/cdk-3.0-minishift-windows-amd64.exe?workflow=direct",
-        "url": "http://cdk-builds.usersys.redhat.com/builds/weekly/09-May-2017.cdk-3.0.0/windows-amd64/minishift.exe",
-        "filename": "minishift_3_0_0_GA.exe",
-        "sha256sum": "4f51b5b6bc8fc93bda5d25f5f58f213a8165b6c0e0f2b77dbb53ae6da4966068",
-        "version": "3.0.0.GA"
+        "dmUrl": "https://developers.redhat.com/download-manager/jdf/file/cdk-3.0.1-minishift-windows-amd64.exe?workflow=direct",
```
dgolovin: Because CDk version is going to be 3.0.0-2, link in download manager will probably be different 

https://developers.redhat.com/download-manager/jdf/file/cdk-3.0.0**-2**-minishift-windows-amd64.exe?workflow=direct

If new CDK buiild will use the same link it will break DevSuite Online installer, because sha256 for the file would be different and CDK download would fail.

Still in progress to clear that with CDK team.

PR #759, opened 2017-06-21
Fix system test launch for bundled installer

jrichter1:
- unpacking now works properly with bundled installer (added a switch for it)
- added the bundle folder as a chrome argument to protractor (but it keeps prepending dashes, so that's why there is the replace part - but only if the string actually starts with '--')


jrichter1:
test this

jrichter1:
test this

PR #770, opened 2017-07-07
Fix promises chaining in JBoss EAP installer

dgolovin:


dgolovin:
Covered by https://github.com/redhat-developer-tooling/developer-platform-install/pull/767 which eliminates creation of .properties file with additional variables.

PR #772, opened 2017-07-10
JBDS-4455 Update 7zip config for bundled installer on build

jrichter1:
so that bundled installer may work with different file names of the inner archive


jrichter1:
Just fyi, this affects all the builds with build date/build number, it will also eventually affect builds with versions beyond 2.0.0

PR #781, opened 2017-07-18
JBDS-4463 Bundle Installer fails to install fuse tools

dgolovin:
Fix add extra check for bundled files and init downloadedFile location
with bundled file.


jrichter1:
test this

PR #782, opened 2017-07-18
JBDS-4464 Fuse Platform does not get added to runtime detection

dgolovin:
Fix attaches runtime detection configuration to installation
of fuse tools.


jrichter1:
test this

jrichter1:
And what happens if you install fuse platform and devstudio without the fuse tooling?

PR #788, opened 2017-07-20
Update devstudio to GA

jrichter1:
to appease the ui tests


jrichter1:
It looks like GA nightlies have different file name format, but the content.json does not reflect that. I'll try using the latest-installer symlink instead, since it looks like the format is the same for all milestones, including GA.

PR #790, opened 2017-07-20
JBDS-4469 Some servers added to runtime detection only with fuse tools

dgolovin:
This fix:
1. Removes not deeded listeners for cmponents required OpenJDK
   to be installed first, because install ordering is already taking
   care of it
2. Fix unit test errors under macOS

### DIFF
--- a/browser/model/jbossfuse.js
+++ b/browser/model/jbossfuse.js
@@ -73,14 +73,14 @@ class FusePlatformInstall extends InstallableItem {
     return Promise.resolve().then(()=> {
       return installer.writeFile(this.installConfigFile, installGenerator(this.installerDataSvc.fuseplatformDir()));
     }).then(()=> {
-        return this.headlessEapInstall(installer);
+      return this.headlessEapInstall(installer);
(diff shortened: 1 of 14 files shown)


browser/model/jbossfuse.js
```diff
@@ -73,60 +73,28 @@ class FusePlatformInstall extends InstallableItem {
...
-        devstudio.configureRuntimeDetection('fuse-platform-on-eap', this.installerDataSvc.fuseplatformDir());
-      } else {
-        this.ipcRenderer.on('installComplete', (event, arg)=> {
-          if(arg == 'fusetools') {
+        return this.headlessInstall(installer);
+    }).then(()=> {
+      this.ipcRenderer.on('installComplete', (event, arg)=> {
+        if(arg == 'all') {
+          let devstudio = this.installerDataSvc.getInstallable('devstudio');
+          if(devstudio.installed()) {
```
dgolovin: This is obviously mistake, should be just 
```  javaScript
devstudio.installed
```

PR #791, opened 2017-07-21
JBDS-4444 Move Account page after Confirmation page

dgolovin:
PR for https://issues.jboss.org/browse/JBDS-4444.


jrichter1:
I like it, one suggestion though. Maybe we could mark the components that require a subscription/login? Then in the top info it could say something like:
"Select the components to install, components marked with (whatever) require RHD login"
WDYT?

dgolovin:
I like your thinking, it might be part of https://issues.jboss.org/browse/JBDS-4335 solution. Lets wait for UX review on this fix and then decide how we should move forward,

PR #803, opened 2017-07-28
JBDS-4484 Fuse Tooling, Fuse and JBoss EAP should not be selected

dgolovin:


browser/model/devstudio.js
```diff
@@ -11,7 +11,7 @@ import JdkInstall from './jdk-install';
 import Platform from '../services/platform';
 
 class DevstudioInstall extends InstallableItem {
-  constructor(keyName, installerDataSvc, targetFolderName, downloadUrl, fileName, sha256sum, additionalLocations, additionalIus, useDownload) {
+  constructor(keyName, installerDataSvc, targetFolderName, downloadUrl, fileName, sha256sum, additionalLocations, additionalIus, useDownload, defaultOption = 'install') {
```
jrichter1: Wouldn't it be nicer if defaultOption was moved to InstallableItem and loaded from its requirement? With install being the default if the requirement has none specified of course.

dgolovin: yep, that is much nicer.

PR #805, opened 2017-07-28
JBDS-4272 Create symlinks in /usr/local/bin running script with admin…

dgolovin:
… rights


jrichter1:
Well, it works, but it turns the installation into a password prompt parade. It's still better than failing on lack of permissions, but some less cumbersome way had better be on the horizon.

dgolovin:
Right, that solution makes sense for now, better ask twice than fail. We'll find better one I mentioned in [another JIRA](https://issues.jboss.org/browse/JBDS-4121?focusedCommentId=13438281&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-13438281)

PR #807, opened 2017-07-28
New gulp task to run tests one by one

dgolovin:


gulp-tasks/tests.js
```diff
@@ -8,6 +8,46 @@ var yargs = require('yargs');
 var buildFolder = path.join('dist', process.platform + '-' + process.arch);
 
 module.exports = function(gulp) {
+
+  function createMocha() {
+    return new mochaa({
```
jrichter1: Looks like a typo. Do you intend to use this function by the way? :)

dgolovin: oh, thanks! I fixed that as different commit. This function is not needed, just leftover from my experiments.

PR #808, opened 2017-07-29
Load components into confirm page using templates

dgolovin:


jrichter1:
The only issue I've found in this state is that virtualbox won't go away when hyper-v is detected properly.

PR #829, opened 2017-08-10
JBDS-4482 Create 'About' dialog to show most relevant information about installer

sudhirverma:


browser/app.css
```diff
@@ -21,7 +21,7 @@ html {
   z-index: 6;
 }
 
-footer {
+.footer {
```
dgolovin: Changing footer class breaks confirmation page layout

image

main/index.js
```diff
@@ -102,3 +106,26 @@ app.on('ready', function() {
     }
   });
 });
+
+function openAboutWindow() {
```
dgolovin: This is not right place to declare a function, index.js is entry point for Installer application. Windows related menu is defined in browsr/bootstrap.js file.

main/menu.js
```diff
@@ -1,35 +1,12 @@
 'use strict';
```
dgolovin: This file is also used on macOS to create macOS specific menu, this menu is not visible under Windows.

browser/about.html
```diff
@@ -0,0 +1,22 @@
...
+  <head>
+    <title>About</title>
+    <link rel="stylesheet" href="app.css">
+  </head>
+  <body>
+      <img class="image" style="margin-right:90px;margin-bottom:27px;" width="220px" src="images/RHDP_logo_RGB_default.svg"/>
+      <section class="welcome-panel">
+        <p>For help, please visit</p>
+        <ul>
+          <li><a href="https://access.redhat.com/documentation/en/red-hat-development-suite/">Documentation</a>.</li>
```
dgolovin: These links loads referenced documents right in about window like this:

image

dgolovin:
Here is PR https://github.com/redhat-developer-tooling/developer-platform-install/pull/841 that shows some of the issues fixed.

browser/pages/about/controller.js
```diff
@@ -10,12 +10,16 @@ class AboutController {
...
-  next() {
-    this.router.go('location');
+  documentation() {
+    this.electron.shell.openExternal('https://access.redhat.com/documentation/en/red-hat-development-suite/');
   }
 
-  openDevSuiteOverview() {
-    this.electron.shell.openExternal('https://developers.redhat.com');
+  release() {
+    this.electron.shell.openExternal('https://access.redhat.com/documentation/en-us/red_hat_development_suite/1.4/html/release_notes_and_known_issues/');
```
dgolovin: The installer version should not be hardcoded into the link. Use https://www.npmjs.com/package/semver module functions:
* major(version)
* minor(version)
to generate right url like
``` javascript
get shortVersion() {
  return `${semver.major($scope.version)}.${semver($scope.version).minor()}`;
}

release() {
  this.electron.shell.openExternal(`https://access.redhat.com/documentation/en-us/red_hat_development_suite/${this.shortVersion}/html/release_notes_and_known_issues/`);
}
```

browser/pages/about/controller.js
```diff
@@ -10,12 +10,16 @@ class AboutController {
...
+  }
+
+  report() {
+    this.electron.shell.openExternal('https://access.redhat.com/documentation/en-us/red_hat_development_suite/1.4/html/installation_guide/troubleshooting#reporting_an_issue');
```
dgolovin: use the same approach mentioned above.

browser/pages/about/about.html
```diff
@@ -1,59 +1,14 @@
...
-  </div>
-
-</form>
+  <form>
+    <img class="image" style="margin-right:90px;margin-bottom:27px;" width="220px" src="images/logo_RHD_RGB_default.svg"/>
+    <span class="version-panel">Version: {{version}}</span>
+      <section class="welcome-panel">
+        <p>For help, please visit</p>
+        <ul>
+          <li><a href="" ng-click="aboutCtrl.documentation()">Documentation</a>.</li>
```
dgolovin: Dots are not required here.

browser/bootstrap.js
```diff
@@ -39,18 +66,19 @@ const help = new MenuItem({
...
-  menu.popup(remote.getCurrentWindow());
-}, false);
-
 function restoreMenu() {
   menu = new Menu();
   menu.append(toggleDevToolsItem);
   menu.append(inspectElementItem);
   menu.append(separator);
   menu.append(help);
+  menu.append(aboutItem);
```
dgolovin: ... and there should be separator between Help and About items.

sudhirverma:
test this

sudhirverma:
test this

browser/bootstrap.js
```diff
@@ -27,6 +28,32 @@ const inspectElementItem = new MenuItem({
   }
 });
 
+function openAboutWindow() {
+
+  let aboutWindow = new BrowserWindow({
```
dgolovin: Should be:
``` JavaScript
let aboutWindow = new BrowserWindow({
    parent: remote.getCurrentWindow(),
    modal: true,
    useContentSize: true,
    width: 565,
    height: 355,
    'autoHideMenuBar': true,
    resizable: false,
    show: false
  });
```
to keep original background size.

That would get us even closer to original design, but still vertical spacing between text requires some adjustments.

image

sudhirverma:
aboutpage

jrichter1:
Not a fan of this solution, especially when cmd+w shows greyed out in the main menu (still works though) until the modal window is closed. I would like it to be something within the app itself that I could click :)

Obviously, the nicest way would be to have some sort of close button as part of the dialog. We could also close the dialog when the main window regains focus (i.e. when the user clicks back into it), but like the current solution it might not be totally apparent thing to do.

Anyway, I'll let you merge this if you pledge to find a better way in this release.

jrichter1:
test this

sudhirverma:
test this

PR #839, opened 2017-08-22
JBDS-4410 Add download size for packages

mohitsuman:
https://issues.jboss.org/browse/JBDS-4410

Added the download size information with the version name. Mockup is attached to the jira.

### DIFF
--- a/browser/directives/componentPanel.html
+++ b/browser/directives/componentPanel.html
@@ -26,7 +26,7 @@
           </div>
         </div>
         <span id="{{item.keyName}}-name" class="product-name">{{item.productName}}</span>
-        <span id="{{item.keyName}}-version" class="product-info">{{item.version}}&nbsp;|&nbsp;{{item.size}}</span>
+        <span id="{{item.keyName}}-version" class="product-info">{{item.version}}</span><span class="product-info">|</span><span id="{{item.keyName}}-size" class="product-info">{{item.size}}</span>
         <div id="{{item.keyName}}-description">{{item.productDesc}}</div>
       </div>
     </div>
--- a/browser/directives/componentPanel.html
+++ b/browser/directives/componentPanel.html
@@ -25,7 +25,8 @@
             </div>
           </div>
         </div>
-        <span id="{{item.keyName}}-name" class="product-name">{{item.productName}}</span><span id="{{item.keyName}}-version" class="product-version">{{item.version}}</span>
+        <span id="{{item.keyName}}-name" class="product-name">{{item.productName}}</span>
+        <span id="{{item.keyName}}-version" class="product-version">{{item.version}}</span>&nbsp;|&nbsp;<span>{{item.size}}</span>
         <div id="{{item.keyName}}-description">{{item.productDesc}}</div>
       </div>
     </div>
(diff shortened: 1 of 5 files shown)


dgolovin:
test this

dgolovin:
test this

dgolovin:
test this

dgolovin:
test this

requirements.json
```diff
@@ -15,7 +15,8 @@
     "sha256sum": "59560c89b76f57ee72d9bd20bac047146f6f369fb42fba94ebaccd65c71209f5",
     "version": "11.0.0.GA",
     "useDownload": true,
-    "requires": ["jdk"]
+    "requires": ["jdk"],
+    "size": "645.38 MB"
```
dgolovin: We should agree on units we use in size field (lets say Kilobytes) and put only number into this field. That would simplify total download computation. On UI side I would use angular filter and transform the number with humanize module to readable value.

Filter would look like:

``` JavaScript
module.filter('humanReadableSize', function() {
    return function(input) {
        return humanize.filesize(input*1000); //be careful with with kilobyte (x1000) and Kibibyte (x1024).
    }
});
```

then on template side:

``` html
{{ item.size | humanReadableSize }}
```

browser/model/installable-item.js
```diff
@@ -72,6 +72,10 @@ class InstallableItem {
     return this.productVersion;
   }
 
+  getProductSize() {
```
mohitsuman: I have added the function as to calculate total download size which we will show. Do you want to get this code in that PR. I will make that change

dgolovin: It is up to you, but if you add new function, please, add unit test for that as well.

mohitsuman: Removed the function as of now. Will use this for calculating total download size.

dgolovin:
test this

mohitsuman:
test this

dgolovin:
test this

browser/directives/componentPanel.html
```diff
@@ -25,7 +25,8 @@
             </div>
           </div>
         </div>
-        <span id="{{item.keyName}}-name" class="product-name">{{item.productName}}</span><span id="{{item.keyName}}-version" class="product-version">{{item.version}}</span>
+        <span id="{{item.keyName}}-name" class="product-name">{{item.productName}}</span>
+        <span id="{{item.keyName}}-version" class="product-info">{{item.version}}&nbsp;|&nbsp;{{item.size}}</span>
```
dgolovin: UI test errors related to change n this line. adding span tag per text element w/ or w/o id attribute 

``` html
<span id="{{item.keyName}}-version" class="product-info">{{item.version}}</span><span class="product-info">|</span><span id="{{item.keyName}}-download-size" class="product-info">{{item.size}}</span>
```

fixes the UI test errors.

jrichter1:
Technical question: how do you intend to keep these values fresh? (Because I know someone will forget eventually)
Maybe a build time script? Plus we could use this value instead of calculating it again when downloads start. Doesn't need to be a part of this PR though.

mohitsuman:
test this

dgolovin:
@jrichter1 the same way we update devstudio info with latest nightly build available. That is different issue to address later.

PR #850, opened 2017-08-29
Make failed download progress bar look like a proper failure

jrichter1:
It should suggest a failure just like the failed installation. Hence it should look like a failure, including the new fancy icon and colours.


jrichter1:
test this

PR #859, opened 2017-09-01
Fix download size

mohitsuman:
There were difference in the download size calculated for a component. I have made the changes accordingly for every component. Now the size is stored in bytes(binary).


mohitsuman:
test this

PR #871, opened 2017-09-08
JBDS-4310 Verify what resources is already downloaded

dgolovin:
Fix adds check for downloaded resources based on file size.
SHA256 sum verification is done later during installation.
Checking file size is much faster than calculating SHA256
for every downloaded file.


jrichter1:
test this

jrichter1:
test this

jrichter1:
test this

PR #873, opened 2017-09-11
fix back button alignment in footer

mohitsuman:
image

PR #884, opened 2017-09-19
	fix back button alignment in footer

dgolovin:


jrichter1:
Is this supposed to be a replacement for #873? Please close the non-relevant one so I know which PR to nag at :)

PR #913, opened 2017-10-11
Fix #911 Add Eclipse Che Installer

### LINKED ISSUE #911: Add Eclipse Che installer
Add Eclipse Che to the list of components available for installation assuming we are going to achieve this with CDK add-on. It means Che installer will require CDK and enable CDK add-on during installation.  
- [X]     Create empty installer and add Che into list of components (https://github.com/redhat-developer-tooling/developer-platform-install/pull/913)
- [X]     Implement Che add-on (https://github.com/minishift/minishift-addons/pull/43)
- [x]     Install Che add-on in CDK
- [x]     Implement installer for Che that enables 'che' add on on CDKE
- [ ]     Add link to documentation how to start and use Eclipse Che

dgolovin:
image

### DIFF
--- a/browser/model/che.js
+++ b/browser/model/che.js
@@ -1,0 +1,44 @@
+'use strict';
+
+import path from 'path';
+import InstallableItem from './installable-item';
+import Installer from './helpers/installer';
+import Platform from '../services/platform';
+
+class CheInstall extends InstallableItem {
+  constructor(installerDataSvc, url) {
+    super(CheInstall.KEY, url, '', '', installerDataSvc, false);
+    this.addOption('install', this.version, '', false);
+  }
+
+  static get KEY() {
+    return 'cdkcheaddon';
+  }
+
+  installAfterRequirements(progress, success, failure) {
+    progress.setStatus('Installing');
+    let installer = new Installer(this.keyName, progress, success, failure);
+    return Promise.resolve().then(()=>{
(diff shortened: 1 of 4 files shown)


dgolovin:
test this

browser/model/che.js
```diff
@@ -0,0 +1,44 @@
...
+
+  static get KEY() {
+    return 'che';
+  }
+
+  installAfterRequirements(progress, success, failure) {
+    progress.setStatus('Installing');
+    let installer = new Installer(this.keyName, progress, success, failure);
+    return Promise.resolve().then(()=>{
+      return installer.exec('minishift config set memory 5GB && minishift addon enable che', this.createEnvironment());
```
budhrg: @LalatenduMohanty looks like we are enabling during installation. I think we need to just set memory when user click to install Che. and point to documentation to apply che.

budhrg: /cc @dgolovin

dgolovin: @budhrg @LalatenduMohanty Why would we call it Eclipse Che Installer then?

LalatenduMohanty: @dgolovin is right. Looks good to me. @budhrg the idea of installer is to make the process super easy for user. So this is the right thing to do as Che would be a separate entity in installer.

budhrg:
Need rebase.

PR #923, opened 2017-10-17
Replace 'download size' to 'amount' throughout install controller

dgolovin:
Current progress holder is intended to report download progress
only. This commit makes it generic and replaces 'download size'
related names to just 'amount' throughout installation page controller.


dgolovin:
test this

PR #931, opened 2017-10-19
Fix #930 Hyper-V Administrators group localization issue

### LINKED ISSUE #930: Possible localization issue when adding user to Hyper-V Administrators group
https://github.com/redhat-developer-tooling/developer-platform-install/blob/5ec2f16dab28877f0d4cac09b1f6261e1c7c215b/browser/model/cdk.js#L48

this line refers to the group by string, instead of a security ID. This has caused issues on Minishift, see: https://github.com/minishift/minishift/issues/1541

When using a different locale, like pt-BR this group is called: "Administradores do Hyper-V"

dgolovin:
Local group names are localized on windows and therefore to support windows installations with different locale CDK installer cannot use 'Hyper-V Administrators' group name when adding current user into it using `net localgroup` command. First it should get localized name by SID (security identifier) for the group using powershell script.

### DIFF
--- a/browser/model/cdk.js
+++ b/browser/model/cdk.js
@@ -44,11 +44,12 @@ class CDKInstall extends InstallableItem {
       let hv = this.installerDataSvc.getInstallable('hyperv');
       if (hv && hv.hasOption('detected')) {
         driverName = 'hyperv';
-        return installer.exec(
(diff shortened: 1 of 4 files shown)


dgolovin:
test this

Issue #929, opened 2017-10-18
Update kompose to v1.3.0 


dgolovin:
Update kompose to latest available release [1.3.0](https://github.com/kubernetes/kompose/releases/tag/v1.3.0)

PR #940, opened 2017-10-23
Update kompose to v1.3.0 #929

sudhirverma:


dgolovin:
@cdrage FYI: we are going to release kompose 1.3.0 instead of 1.4.0 we agreed before. That also implies releasing kompose 1.3.0 to devtools channel for RHEL7.  The reason is 1.4.0 is going to be released to close to the end of the current release train.

cdrage:
@dgolovin @sudhirverma Yup! All good with me. :100:

PR #983, opened 2017-10-28
[WIP]Replace localStorage with node-localstorage

dgolovin:


test/unit/model/cdk-test.js
```diff
@@ -16,6 +16,7 @@ import InstallableItem from 'browser/model/installable-item';
 import child_process from 'child_process';
 import mockfs from 'mock-fs';
 import loadMetadata from 'browser/services/metadata';
+import TockenStore from 'browser/services/credentialManager';
```
mohitsuman: This should be replaced with `TokenStore` as the exported class name.

PR #1026, opened 2017-11-08
Import virtualbox certificate before installation (fix #1013)

### LINKED ISSUE #1013: Investigate accepting certificate for virtualbox drivers without additional user input
If the system does not have the proper oracle certificate in its trusted authorities, virtualbox installation will present a modal dialog asking the user to accept it. Find a way to avoid the dialog and do it automatically.

jrichter1:
Extract the certificate from virtualbox executable and import it before installing. In case somebody does not have it yet, it should no longer prompt for "trusting oracle" when installing vbox drivers.

### DIFF
--- a/browser/model/virtualbox.js
+++ b/browser/model/virtualbox.js
@@ -118,9 +118,13 @@ class VirtualBoxInstallWindows extends VirtualBoxInstall {
 
   installAfterRequirements(progress, success, failure) {
     let installer = new Installer(this.keyName, progress, success, failure);
-    return installer.exec(
-      [`"${this.downloadedFile}"`, '--extract', '-path', `"${this.installerDataSvc.virtualBoxDir()}"`, '--silent'].join(' ')
-    ).then(() => {
+    return this.importCertificate(installer
+    ).catch((error) => {
+      Logger.info(this.keyName + ' - Skipping certificate import due to error');
+      Logger.error(this.keyName + ' - ' + error);
+    }).then(() => {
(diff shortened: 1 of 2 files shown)


browser/model/virtualbox.js
```diff
@@ -118,9 +118,12 @@ class VirtualBoxInstallWindows extends VirtualBoxInstall {
 
   installAfterRequirements(progress, success, failure) {
     let installer = new Installer(this.keyName, progress, success, failure);
-    return installer.exec(
-      [`"${this.downloadedFile}"`, '--extract', '-path', `"${this.installerDataSvc.virtualBoxDir()}"`, '--silent'].join(' ')
-    ).then(() => {
+    return this.importCertificate(installer
+    ).catch(() => {
+      // continue even if certificate fails to import automatically
```
dgolovin: I don't like empty catch blocks, lets at least log warning here.

browser/model/virtualbox.js
```diff
@@ -136,6 +139,19 @@ class VirtualBoxInstallWindows extends VirtualBoxInstall {
     });
   }
 
+  importCertificate(installer) {
+    let command = [
```
dgolovin: There should be factory method for command.

test/unit/model/virtualbox-test.js
```diff
@@ -162,6 +162,27 @@ describe('Virtualbox installer', function() {
         installer.ipcRenderer = {on: function() {}};
       });
 
+      it('should import the certificate from the executable', function() {
+        sandbox.stub(child_process, 'execFile').yields('done');
+
+        let command = [
```
dgolovin: Use command factory method here to avoid copy-paste.

jrichter1:
Issues have been addressed, hopefully ;)

Issue #1020, opened 2017-11-06
JavaScript error occurred in main process on macOS close window keys combination


dgolovin:
![image](https://user-images.githubusercontent.com/620330/32465025-ec42b8a2-c2f6-11e7-8f90-e00b562b3dc0.png)

PR #1035, opened 2017-11-13
JavaScript error occurred in main process on macOS close window keys combination #1020

sudhirverma:


sudhirverma:
test this

dgolovin:
test this

sudhirverma:
test this

main/index.js
```diff
@@ -49,7 +49,10 @@ app.on('quit', function(event, exitCode) {
 // initialization and is ready to create browser windows.
 app.on('ready', function() {
   globalShortcut.register('CmdOrCtrl+W',()=>{
```
dgolovin: Is there should be a code that removes global shortcut when installer app is closed? What happens if you run installer with this fix, then exit it and then try to use Cmd+W in any other window?

Issue #972, opened 2017-10-27
Uninstall script requires user input when uninstalling full installation


dgolovin:
![image](https://user-images.githubusercontent.com/620330/32091856-e201087a-baaa-11e7-9f87-6641e90ac319.png)

PR #1059, opened 2017-11-24
Uninstall script requires user input when uninstalling full installation #972

sudhirverma:


uninstaller/uninstall-helper.ps1
```diff
@@ -60,7 +60,7 @@ if ($subfolders.Length -gt 0) {
   foreach ($item in $subfolders) {
     robocopy "$folder\..\temp" "$item" /purge | Out-Null
   }
-  Get-ChildItem "$folder\.." -Recurse | Remove-Item -Force
+  Get-ChildItem "$folder\.." -Recurse | Remove-Item -Force -recurse
```
dgolovin: Why not just one liner:

``` powershell
Remove-Item -path "$folder\.." -force -recurse
```

dgolovin:
Fixed in #1068.

Issue #1004, opened 2017-10-31
Change default installation folder on windows to %PROGRAMFILES%\Development Suite


dgolovin:
Instead of using hardcoded location for program files as `C:\Program Files` it should be %PROGRAMFILES% env variable on windows.

dgolovin:
Let's keep the folder name the same as it was for now, until we confirm new name.

PR #1060, opened 2017-11-24
Change default installation folder on windows to %PROGRAMFILES%\Development Suite #1004

sudhirverma:


browser/services/platform.js
```diff
@@ -50,6 +50,16 @@ class Platform {
     });
   }
 
+  static getProgramFilePath() {
+    return Platform.identify({
+      win32: function() {
+        return pify(child_process.exec)('echo %PROGRAMFILES%').then((stdout)=>{
```
dgolovin: Yo can get env variable value just from `process.env.PROGRAMFILES`, no need to run async command.

sudhirverma:
test this

dgolovin:
Replaced with https://github.com/redhat-developer-tooling/developer-platform-install/pull/1069 which address unit test errors when running on linux.

PR #1062, opened 2017-11-27
Fix css changes and add patternfly changes

### LINKED ISSUE #1036: Use patternfly components throughout the installer app
Based on the guidelines and uniformity across different apps, use of `Pattern Fly` is recommended. Will be using the features for the UX components in accordance with that.

mohitsuman:
Fix #1036:

1. Change the use of components to use `patternfly` and then apply custom css.
2. Update the `grid column` style of css.
3. Add UX for login for account.html page.

For row elements of selection page, added border to distinguish between them based on cursor movement. Attached video.

giphy


dgolovin:
test this

dgolovin:
image
Checkbox is not in the middle as it was(?) before.

browser/app.css
```diff
@@ -937,4 +931,19 @@ footer {
     margin-right: 10px;
     float: right;
   }
+
+  .panel-group-items .panel:not(.zero-border):hover {
+    border-left: 3px solid #c00;
```
dgolovin: border-left with 3px wide makes component card moves right 1px when mouse hovers over. We should ether have it 2px, or figure out how to make card stay on the same place when mouse hoovers over it.

dgolovin:
All not primary buttons (Back and Cancel) should look like secondary buttons from patternfly examples.

browser/app.css
mohitsuman: Fixed in https://github.com/redhat-developer-tooling/developer-platform-install/pull/1062/commits/6c04fe5849b2cc6e6cba9eb6257d1a40212922e4

mohitsuman:
The checkbox is also aligned at the centre with the change in padding.  
https://github.com/redhat-developer-tooling/developer-platform-install/pull/1062/commits/6c04fe5849b2cc6e6cba9eb6257d1a40212922e4#diff-7cc3025cbe6f198e4624ce63d1c75636R597

PR #1067, opened 2017-11-28
Remove transpiled folder before running transpiler again

dgolovin:
This fix is for the case when some files are removed,
to avoid running app with deleted files preserved in
transpiled folder


dgolovin:
test this

dgolovin:
test this

PR #1071, opened 2017-11-28
[wip]Queue operations with progress

### LINKED ISSUE #995: Total download size calculation  should support components with multiple files to download
Current implementation works correctly only in case of 1 to 0..1 relation between component declaration and binary file to download. Fuse for EAP does not fit in there, because it has two resources to download Fuse EAP insatller and EAP installer itself (also see #909). It always shown as 'Selected to download' on confirmation page
image

### LINKED ISSUE #1012: Make use of the progress bar when verifying checksums
The verification may take a notable amount of time for multiple files. Use the progress bar to at least indicate how many components are done/left.

jrichter1:
This turned out way bigger than originally intended (to fix #1012). On the upside, it now fixes #995 as well.
Changelog:
 - split checkAndDownload into check and download (if check is successful, download is not even called)
 - made all checks done per file for each component
 - InstallerDataService takes care of delegating to each component, the install controller now mostly just uses that
 - the flow is now 'verify existing => download (and check) => install' using events to advance to next step
 - all steps now use the progress bar
 - post download checks are now queued asynchronously to the download, and if some remain after download is done, they take over the progress bar

Take it for a spin please, and once again sorry for how big it turned out to be - it just didn't seem to work in smaller pieces.

### DIFF
--- a/browser/model/devstudio.js
+++ b/browser/model/devstudio.js
@@ -10,7 +10,7 @@ import Logger from '../services/logger';
 import Platform from '../services/platform';
 
 class DevstudioInstall extends InstallableItem {
-  constructor(keyName, installerDataSvc, targetFolderName, downloadUrl, fileName, sha256sum, additionalLocations, additionalIus, useDownload) {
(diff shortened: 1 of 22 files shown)


jrichter1:
test this

PR #1073, opened 2017-11-28
Revert to use gulp-symlink instead of vinyl-fs

dgolovin:
Reverts back to using gulp-symlink because it does not require elevated session to work.


mohitsuman:
@dgolovin The reason to use `vinyl-fs` was because `gulp-symlink` is deprecated now. The official document mentions to use vinyl-fs. 
https://github.com/ben-eb/gulp-symlink#deprecation-notice

What exact change do we need to make in the new package for the elevated session to work, so that we do not use the deprecated one ?

dgolovin:
https://github.com/ben-eb/gulp-symlink#deprecation-notice say: "gulp 4 adds built-in symlinks to the public API".   We should migrate to gulp 4.

mohitsuman:
gulp 4 is not officially released and it is still in alpha phase from last one year. And as we will be migrating to webpack in the future, gulp 4 won't be an option.

dgolovin:
I fail to see how using webpack stops us from using gulp of any version.
Another option to check is to use fs.symlink directly, but for now I'd rather use deprecated gulp-symlink that run gulp in elevated terminal on Windows.

PR #1076, opened 2017-11-29
Remove unwanted css and add note for close window on macOS

### LINKED ISSUE #1075: Fix about page and welcome page css
- Update the css classes and remove the unwanted classes for welcome and about page.

- Add **Note** for closing the window `About` modal on macOS.

mohitsuman:
Fix #1075 

screen shot 2017-11-29 at 7 13 56 pm

### DIFF
--- a/browser/about.html
+++ b/browser/about.html
@@ -7,6 +7,6 @@
     <link rel="stylesheet" href="app.css">
     <script src="./about.js"></script>
   </head>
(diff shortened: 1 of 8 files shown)


mohitsuman:
@jrichter1 I have added the note at the `About` modal for macOS. This will give the user an idea of how to close the modal. As we have discussed, Win and Linux have close buttons/icon which make it self explanatory. 

As macOS as the modals are added as sheets attached to the parent window, there is no close button. So for now we have added the `Note` info at the bottom.

PR #1080, opened 2017-11-30
Fix usage of 'useDownload' attribute

dgolovin:
'useDownload' originally means 'No download required' because this installer has no files to download.


jrichter1:
test this

Issue #1066, opened 2017-11-27
Select component for installation with a mouse click inside card


dgolovin:
Currently selection is done by clicking directly on checkbox at the left-hand side.

jrichter1:
A silly observation: you can select/deselect components by clicking anywhere but on the checkbox itself.

dgolovin:
nice catch 👍

PR #1085, opened 2017-12-01
Select component for installation with a mouse click inside card #1066

sudhirverma:


browser/app.css
```diff
@@ -504,6 +511,15 @@ footer {
     border-width: 2px
   }
 
+  .panel-border {
+    margin-bottom: 20px;
+    background-color: rgba(93, 77, 77, 0.20);
```
mohitsuman: Convert the value to hex code.

browser/app.css
```diff
@@ -504,6 +511,15 @@ footer {
...
+    border: 1px solid transparent;
+    border-radius: 1px;
+    height: 74px;
+    box-shadow: 0 1px 1px rgba(0, 0, 0, 0.05);
```
mohitsuman: use hex value for the color code.

mohitsuman:
@sudhirverma The changes you have applied override the previous component layout.

- The dotted border layout does not for already installed component. See attached image.
screen shot 2017-12-01 at 7 30 08 pm

- The component border when we hover is also missing as in the case we had previously. 

- The font for component description is also bold, which should be normal.

Please fix these changes.

browser/app.css
mohitsuman: use hex value for the color code.

browser/app.css
```diff
@@ -10,6 +10,13 @@ header {
   text-align: center;
 }
 
+.label-panel {
+  display: inline-block;
+  max-width: 100%;
+  margin-bottom: 5px;
+  width: 1439px;
```
mohitsuman: We do not need width of so much pixels. Instead use `width: 100%` and remove `max-width` parameter.

Why are we using `margin-bottom` here. It will not add anything as it is already inside div container. So we can get rid of this also.

dgolovin:
Resolved in https://github.com/redhat-developer-tooling/developer-platform-install/pull/1087.

Issue #1010, opened 2017-11-02
Uninstaller fails to run from Windows7 Control Panel


dgolovin:

dgolovin:
Fixed with https://github.com/redhat-developer-tooling/developer-platform-install/pull/1096

PR #1096, opened 2017-12-06
Uninstaller fails to run from Windows7 Control Panel #1010

sudhirverma:


uninstaller/uninstall-helper.ps1
```diff
@@ -72,7 +72,12 @@ $pathFolders | foreach {
 [string] $delimitedFolders = $folderList -Join ';'
 [Environment]::SetEnvironmentVariable("Path", $delimitedFolders, "User")
 
-Remove-Item -Path "HKLM:\SOFTWARE\WOW6432Node\Microsoft\Windows\CurrentVersion\Uninstall\DevelopmentSuite$timeStamp"
```
dgolovin: It seems that the only difference between what work and what does not is `-Recurse` parameter.

uninstaller/uninstall-helper.ps1
```diff
@@ -72,7 +72,12 @@ $pathFolders | foreach {
...
+for ($i=0; $i -le $key.Count; $i++) {
```
dgolovin: This would be definitely data loss in case there are several devsuite installs on the same machine.

uninstaller/uninstall-helper.ps1
sudhirverma: every time when we install it generate random key in `"HKLM:\SOFTWARE\WOW6432Node\Microsoft\Windows\CurrentVersion\Uninstall\"` that is why I am matching the key name and deleting it.

dgolovin: The key name formed from two parts DevelopmentSuite + $timeStamp from line [#5](https://github.com/redhat-developer-tooling/developer-platform-install/pull/1096/files#diff-61d71e285c0cfd8483e06293eda5583aL6). Why iterate when you can just have it as `"HKLM:\SOFTWARE\WOW6432Node\Microsoft\Windows\CurrentVersion\Uninstall\DevelopmentSuite$timeStamp"`?

uninstaller/uninstall-helper.ps1
dgolovin: Where is $key variable coming from?

dgolovin: This would not work for scenario:
1. Install DevSuite into `c:\Program Files\DevelopmentSuite';
2. Install DevSuite into `c:\Program Files\DevelopmentSuiteAgain';
3. You have two entries in control panel;
4. Uninstall one install;
5. Check if another one is still there. If both are gone you have a problem.

sudhirverma: ooh I  forgot to write `$key = Get-ChildItem "HKLM:\SOFTWARE\WOW6432Node\Microsoft\Windows\CurrentVersion\Uninstall" -Name`  code

uninstaller/uninstall-helper.ps1
```diff
@@ -55,7 +55,16 @@ $targetFolder = [System.IO.Path]::GetFullPath((Join-Path ($folder) '..'))
 
 echo 'Removing installation folder'
 
-Remove-Item -path "\\?\$targetFolder" -Force -Recurse
+$major = $PSVersionTable.PSVersion.Major
+$minor = $PSVersionTable.PSVersion.Minor
+
+if ("$major.$minor" -gt 5.0) {
```
dgolovin: This is not working as expected:
``` shell
PS C:\Users\eskimo> "5.1" -gt 5.0
True

PS C:\Users\eskimo> "5.0" -gt 5.0
True

PS C:\Users\eskimo> "4.9" -gt 5.0
False
```

Should be 

``` shell
if ("$major.$minor" -gt "5.0") {
```

so 

``` shell
PS C:\Users\eskimo> "5.1" -gt 5.0
True

PS C:\Users\eskimo> "5.0" -gt 5.0
True

PS C:\Users\eskimo> "4.9" -gt 5.0
False
```

uninstaller/create-uninstaller.ps1
```diff
@@ -2,11 +2,12 @@
 $timeStamp = $args[1]
 $versionString = $args[2]
 
+$powershell = which powershell
```
dgolovin: This does not work for me on Windows 10.  It should be:

``` powershell
$powershell = where.exe powershell
```

It should be `where.exe` not 'where', because `where` is an alias for powershell's `where-object` cmdlet.
Please verify if it works for Windows 7.

sudhirverma: it's working on Windows 7.

PR #1102, opened 2017-12-07
add package-lock to repo

### LINKED ISSUE #1101: update package json to latest and commit package.lock to repo
As per the new npm modules, `package-lock.json` should be committed in the repo. Upgrading the dependencies to latest and committing the changes,

mohitsuman:
Fix #1101


mohitsuman:
@dgolovin The idea to add package-lock.json in the source repository was from the npm documentation. 
As per the help page of https://docs.npmjs.com/files/package-lock.json, `package-lock.json` is intended to be checked into source control. We may see warning while doing npm install on the command line: `created a lockfile as package-lock.json. You should commit this file.`

It is highly recommended to use package-lock.json which is generated using npm 5.x, as it prevents the "works on my machine" effect. Note: Even upstream projects like fabric8 use the same.

Issue #1082, opened 2017-11-30
Replace hash.js with hasha npm module


dgolovin:
Details are here https://www.npmjs.com/package/hasha
Example is below:
``` javascript
const hasha = require('hasha');
 
// Get the MD5 hash of an image 
hasha.fromFile('unicorn.png', {algorithm: 'md5'}).then(hash => {
    console.log(hash);
    //=> '1abcb33beeb811dca15f0ac3e47b88d9' 
});
```

PR #1119, opened 2017-12-11
Replace hash.js with hasha npm module #1082

sudhirverma:


jrichter1:
Wouldn't it be nicer to remove hash.js all together and replace all calls to it with hasha?

dgolovin:
That was original idea, but then we decided to keep it simple. The reason to keep this file is to skip all the refactoring and fixing unit tests errors throughout the project.

browser/model/helpers/hash.js
```diff
@@ -1,24 +1,10 @@
...
-      }).on('close', function () {
-        var hashstring = hash.digest('hex');
-        done(hashstring);
-        res(hashstring);
-      }).on('error', function(err) {
-        done(undefined, err);
-        rej(err);
-      });
+  SHA256(filename) {
+    return hasha.fromFile(filename, {algorithm: 'sha256'}).then((hash) =>{
```
dgolovin: you can just have 
``` JavaScript
return hasha.fromFile(filename, {algorithm: 'sha256'});
```

PR #1125, opened 2017-12-12
Added few file in .gitignore

sudhirverma:


jrichter1:
I think it better to fix the cause rather than treat the symptoms, #1126 should take care of that :)

Issue #1120, opened 2017-12-12
Make the Select/Clear all buttons only affect the selected channel (?)


jrichter1:
Not quite sure how to approach this, but it feels weird these buttons will affect components you are not looking at. We could try to make them work for the selected channel only, that way what you see is what you get. Opinions?

dgolovin:
Agreed, assigning to @sudhirverma.

dgolovin:
Fixed in https://github.com/redhat-developer-tooling/developer-platform-install/pull/1172.

PR #1166, opened 2017-12-20
Channels navbar UI enhancement

mohitsuman:
Fix #1165 
Making the changes for the vertical navbar in channels UI.

screen shot 2017-12-20 at 2 09 44 am

PR #1172, opened 2017-12-21
Make the Select/Clear all buttons only affect the selected channel  #1120

sudhirverma:


browser/pages/selection/selection.html
```diff
@@ -15,14 +15,14 @@
           <div class="col-sm-4 channels-nav">
               <ul class="nav nav-tabs nav-stacked nav-pills" role="tablist">
                   <li id="{{tab.id}}-tab" ng-class="{'active': selectCtrl.channel_tab === tab.id}" ng-repeat="tab in selectCtrl.channelList">
-                      <a class="btn-lg" ng-click="selectCtrl.channel_tab = tab.id" href="">{{tab.name}}</a>
+                      <a class="btn-lg" ng-init="selectCtrl.componentsInChannel('containerDev')" ng-click="selectCtrl.componentsInChannel(tab.id)" href="">{{tab.name}}</a>
```
mohitsuman: We already have defined `channel_tab` initial value as 'containerDev'. There should not be any need to define again in ng-init.

mohitsuman:
@sudhirverma Why is there a question mark for the PR name ? :)

browser/pages/selection/selection.html
sudhirverma: Thanks mohit I will make the changes :)

sudhirverma: If we remove 'containerDev' from ng-init then we can't select the other channels.

requirements.json
```diff
@@ -23,7 +23,8 @@
       "javaDev": { },
       "fuseDev": { },
       "rhProds": { },
-      "devstudio": { }
+      "devstudio": { },
+      "all": { }
```
mohitsuman: Why are we adding all object in all the components. Any specific reason for this?

browser/pages/selection/selection.html
mohitsuman: ng-init is not the right way to use the scope value. It is not recommended a best practice. So we should not be using it. https://docs.angularjs.org/api/ng/directive/ngInit

requirements.json
sudhirverma: `node.channel.hasOwnProperty(this.channel_tab)` code return true if it's find oblect. When we select `All` channel `node.channel.hasOwnProperty(this.channel_tab)`  return false because it's not able to find the object that is why I have added the all object in all components.

mohitsuman: This was working before when we do not use the `all` object. We need to find a way to use  without using `hasOwnProperty`.

sudhirverma:
test this

browser/pages/selection/selection.html
sudhirverma: when we are using bind it's calling all component twice but when we are using `ngInit`  it calls all component only once, So I think we can use `ngInit` here.

sudhirverma:
test this

dgolovin:
test this

sudhirverma:
test this

dgolovin:
test this

mohitsuman:
Closing this issue as it is fixed by https://github.com/redhat-developer-tooling/developer-platform-install/pull/1166 and pushed to master.

Issue #1170, opened 2017-12-21
DevStudio installer should rename `InstallConfigRecord.xml` to 'InstallConfigRecord-${this.keyName}.xml' after it is finished installation


dgolovin:
That is required to preserve original devstudio InstallConfigRecord.xml and then use shortcuts definitions in it in uninstaller to remove shortcuts, because when 'Fuse Tooling' is installed after devstudio it overrides 'InstallConfigRecord.xml' with new one without shortcuts information in it and uninstaller fails to remove shortcuts in this case.

dgolovin:
Released to master

PR #1176, opened 2017-12-26
DevStudio installer should rename `InstallConfigRecord.xml` to 'InstallConfigRecord-${this.keyName}.xml' after it is finished installation #1170

sudhirverma:


dgolovin:
Included into https://github.com/redhat-developer-tooling/developer-platform-install/pull/1178

PR #1182, opened 2018-01-02
Clean up code for location and confirmation page controllers

dgolovin:


mohitsuman:
test this

browser/pages/confirm/controller.js
```diff
@@ -20,8 +20,8 @@ class ConfirmController {
     for (let value of this.installerDataSvc.allInstallables().values()) {
       if(value.size && value.selectedOption == 'install' && !value.downloaded) {
         for (let file in value.files) {
-          if (!value.files[file].downloaded) {
-            totalDownloadSize += value.files[file].size;
+          if (!file.downloaded) {
```
jrichter1: I don't think you can reference a nested object directly through for...in, since it iterates over properties (keys), not over data. for...of should do that, but plain object are not iterable I'm afraid

dgolovin: That explains all the problems in tests.

jrichter1:
test this

PR #1191, opened 2018-01-05
Update channels description

dgolovin:


dgolovin:
Included into new PR here https://github.com/redhat-developer-tooling/developer-platform-install/pull/1192

PR #1194, opened 2018-01-09
Allow to use installable-item.js as module in requirements.json

dgolovin:
This fix lets to use installable-item.js as module for installers declared in requirements.json. It would just download declared item in cache and installation just calls success method.


dgolovin:
test this

PR #1198, opened 2018-01-09
Fix #1197 Add Red Hat Central to components available for installation

### LINKED ISSUE #1197: Add Red Hat Central repository to components available for installation
This component sholdonly download zipped Red Hat Central repository to DevSuite cache and allow to reuse it to install various central item through DevSuite installer. Component itself would not to do any installation.

dgolovin:


dgolovin:
test this

PR #1200, opened 2018-01-09
 Fix #199 Add AngularJS for Web Tools Eclipse plugin

dgolovin:


dgolovin:
There is still problem with ordering installation.

dgolovin:
test this

PR #1211, opened 2018-01-11
Fix #1187 macOS should warn that the auth dialog will read osascript

### LINKED ISSUE #1187: DevSuite Installer for macOS should warn that the authentication dialog will read osascript
Installation page warning on macOS should mention that authentication dialog will appear several times and read osascript (see also https://github.com/redhat-developer-tooling/developer-platform-install/issues/1131).

dgolovin:


jrichter1:
test this

PR #1250, opened 2018-01-30
Simplify installation ordering

dgolovin:
This fix gets read of installAfter hell in requirements.json. It flatten graph or components and then just set installAfter in simple cycle. Net step would be getting rid of installAfter notion. Installation would just go in cycle for flatten requirements graph.


dgolovin:
Will clean it up. Thanks!

PR #1251, opened 2018-02-01
Create a virtual switch for cdk on hyper-v (fix #1242)

### LINKED ISSUE #1242: Set up a vswitch for CDK to use on Hyper-v
Right now to use fresh hyper-v, the user must set up a vswitch and export a variable to tell CDK to use it. It should be taken care of automatically.

jrichter1:
set up a virtual switch along with the HYPERV_VIRTUAL_SWITCH variable for cdk so the user doesn't need to do it themselves

### DIFF
--- a/browser/model/cdk.js
+++ b/browser/model/cdk.js
@@ -8,6 +8,8 @@ import Installer from './helpers/installer';
 import globby from 'globby';
 import fs from 'fs-extra';
 import pify from 'pify';
+import os from 'os';
+import Util from './helpers/util';
 
 class CDKInstall extends InstallableItem {
   constructor(installerDataSvc, targetFolderName, downloadUrl, fileName, sha256sum) {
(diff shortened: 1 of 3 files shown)


browser/model/cdk.js
```diff
@@ -117,6 +123,22 @@ class CDKInstall extends InstallableItem {
...
   }
+
+  createHypervSwitch() {
+    let commands = [
+      '$ErrorActionPreference = "Stop"',
+      '$switchName = if ($env:HYPERV_VIRTUAL_SWITCH) {$env:HYPERV_VIRTUAL_SWITCH} else {"minishift-switch"}',
+      'try { Get-VMSwitch -Name $switchName }',
+      'catch { $adapterName;',
+      '[array]$adapters = Get-NetAdapter -Physical | Where-Object { $_.status -eq "up" }',
+      'foreach ($adapter in $adapters) {',
```
dgolovin: What if I have more than one adapter?

It seems that we should ether detect adapter with internet connection and use it or change selection page UI to allow some custom settings for installable components.

jrichter1: Well, this script searches for all physical adapters that are "up", which means "enabled" and "connected". Not sure if "connected" means the cable is connected, or there is internet connection. 
If some of them are labeled "Ethernet" (the default name for wired connection), it selects the first one it finds. Otherwise it tries to use any adapter it could find with the above conditions.

What would you like to be improved?

dgolovin: Thanks for explanation, it sounds good to me.

PR #1259, opened 2018-02-07
Replace osascript with sudo prompt

jrichter1:
Use sudo-prompt to get elevated privileges:
 - use actual sudo instead of osascript elevation
 - customizable auth dialog (name and icon match the product, yay)
 - credentials remain active until invalidated (session runs out / sudo -k)
   - unless tty-tickets is enabled, only one dialog will be presented during the installation

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,7 +183,7 @@ of Vagrant and VirtualBox (will download them):
 5. To build both installers in a single step:
 
         npm run dist
-        
+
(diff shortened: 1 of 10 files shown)


mohitsuman:
@jrichter1 How does the auth dialog looks now. Can you add a screenshot of it.

jrichter1:
sudo

PR #1262, opened 2018-02-09
Reworked detection for more speed

jrichter1:
I've been messing around with the detection scripts for hypervisors and a bit for java as well. I've found some alternative ways to detect stuff, which seem to be really fast.

Could you guys try this out and tell me if it works for you, and if so how fast it is? (each component reports the time it took to detect in milliseconds into console)

If all goes well, this might turn into a proper PR :)


dgolovin:
Detection is super fast now!

PR #1271, opened 2018-02-16
Use a different user-agent header for debug/test runs (#fix 1269)

jrichter1:
so that the http requests for testing purposes can be filtered out in the statistics.
Feel free to edit the agent string itself if you don't like it :)

### DIFF
--- a/gulpfile.babel.js
+++ b/gulpfile.babel.js
@@ -23,6 +23,8 @@ var fs = require('fs-extra'),
 require('./gulp-tasks/tests')(gulp);
 require('./gulp-tasks/dist-' + process.platform)(gulp, reqs);
 
+var testAgent = 'RedHatDevelopmentSuiteTestInstaller/' + pjson.version + ' test/' + pjson.version;
+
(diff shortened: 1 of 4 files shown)


gulpfile.babel.js
```diff
@@ -23,6 +23,8 @@ var fs = require('fs-extra'),
 require('./gulp-tasks/tests')(gulp);
 require('./gulp-tasks/dist-' + process.platform)(gulp, reqs);
 
+var testAgent = 'RedHatDevelopmentSuiteInstaller/' + pjson.version + ' test/' + pjson.version;
```
dgolovin: Everything seems to be fine, except this test agent string, it still would be a match for splunk query because regex we use is 
``` regexp
*RedHatDevelopmentSuiteInstaller*
```
Should be 'RedHatDevelopmentSuiteTestInstaller/' so our query would not count it.

jrichter1:
Updated the agent string. The build is red though, need to merge #1276 to fix that.

jrichter1:
test this

Issue #1257, opened 2018-02-06
Add RHAMT CLI to `Java Developmet' channel


dgolovin:
Add installer for RHAMT CLI (see https://developers.redhat.com/download-manager/file/4.0.0/migrationtoolkit-rhamt-cli-4.0.0.offline.zip) to Java Development channel. It requires not just download for zip and unpacking into devsuite target install folder, but also pathcing rh-amt.bat to be able to detect and use installed OpenJDK (currently only Oracle Java is supported).

PR #1273, opened 2018-02-18
[WIP] Add RHAMT CLI to `Java Development' channel #1257

sudhirverma:


browser/model/rhamt.js
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  static get KEY() {
+    return 'rhamt';
+  }
+
+  installAfterRequirements(progress, success, failure) {
+    progress.setStatus('Installing');
+    let installer = new Installer(this.keyName, progress, success, failure);
+    return new Promise((resolve, reject)=> {
```
jrichter1: If simply unzipping is the aim here, why not use what we already have:
https://github.com/redhat-developer-tooling/developer-platform-install/blob/master/browser/model/helpers/installer.js#L78

browser/model/helpers/installer.js
```diff
@@ -100,15 +102,25 @@ class Installer {
...
-        fs.createReadStream(zipFile)
-          .pipe(unzip.Extract({path: extractTo}))
-          .on('close', () => {
-            Logger.info(this.key + ' - Extract ' + zipFile + ' to ' + extractTo + ' SUCCESS');
-            resolve(true);
-          })
-          .on('error', (err) => {
-            Logger.error(this.key + ' - ' + err);
-            reject(err);
+        fs.createReadStream(zipFile).pipe(unzip.Parse())
```
dgolovin: What is wrong with original code?

test/unit/model/rhamt-test.js
```diff
@@ -0,0 +1,132 @@
...
+import RhamtInstall from 'browser/model/rhamt';
+import fs from 'fs';
+import path from 'path';
+import mkdirp from 'mkdirp';
+
+import EventEmitter from 'events';
+
+chai.use(sinonChai);
+
+describe('jbossplaformkaraf nstaller', function() {
```
dgolovin: Should be 'rhamt installer'.

browser/model/helpers/installer.js
jrichter1: At first I thought you could just use the original unzip and then rename the parent folder.
But since this method is no longer used and the new code is clearly the same as https://github.com/redhat-developer-tooling/developer-platform-install/blob/master/browser/model/jbossfusekaraf.js#L27 there might be some use in having it in a lib like this.

browser/model/rhamt.js
```diff
@@ -0,0 +1,42 @@
...
+let unzip = require('unzip-stream');
+
+
+import InstallableItem from './installable-item';
+import Installer from './helpers/installer';
+import Platform from '../services/platform';
+
+class RhamtInstall extends InstallableItem {
+  constructor(installerDataSvc, targetFolderName, downloadUrl, fileName, sha256sum) {
+    super(RhamtInstall.KEY, downloadUrl, fileName, targetFolderName, installerDataSvc, false);
```
dgolovin: last parameter should be `true'
``` JavaScript
super(RhamtInstall.KEY, downloadUrl, fileName, targetFolderName, installerDataSvc, true);
```

requirements.json
```diff
@@ -282,6 +282,32 @@
       }
     }
   },
+  "rhamt": {
+    "name": "Red Hat Application Migration Toolkit",
```
dgolovin: Should be: "Red Hat Application Migration Toolkit CLI"

requirements.json
```diff
@@ -282,6 +282,32 @@
...
+    "description": "",
```
dgolovin: Description should be: "An assembly of open source tools that enables large-scale application migrations and modernizations"

browser/model/rhamt.js
```diff
@@ -0,0 +1,42 @@
...
+    this.addOption('install', this.version, '', true);
+    this.sha256 = sha256sum;
+  }
+
+  static get KEY() {
+    return 'rhamt';
+  }
+
+  installAfterRequirements(progress, success, failure) {
```
dgolovin: Should also add the executables to PATH.

browser/model/helpers/installer.js
dgolovin: Ups, sorry for my confusing comment. I really meant why not to create generic installer from zip files that would unpack archive and then add some specific locations from zip to PATH. We can list files to add to PATH as part of json data.
But I am fine with current solution as well.

browser/model/rhamt.js
```diff
@@ -0,0 +1,63 @@
...
+        fs.readFile(rhamtcli, 'utf8', function (err,data) {
+          if (err) {
+            return console.log(err);
+          }
+          let regeditdata = data.match(/FOR \/F "skip=2 tokens=2.*.B/g)
+          let replacedata = data.replace(/FOR \/F "skip=2 tokens=2.*.B/, ' ');
+          fs.writeFile(rhamtcli, replacedata, 'utf8', function (err) {
+            if (err) return console.log(err);
+          });
+          var openJDK = 'REG QUERY "HKLM\\Software\\JavaSoft\\Java Runtime Environment\\1.8.0_161_1" /v JavaHome'
```
dgolovin: This is not going to work if different OpenJDK version is detected, but it should work with all OpenJDK releases.

sudhirverma:
test this

sudhirverma:
test this

sudhirverma:
test this

sudhirverma:
test this

sudhirverma:
test this

sudhirverma:
test this

browser/model/rhamt.js
dgolovin: We can try different approach here. For windows we would set JAVA_HOME (if it is not set already) to be the location of detected JDK or selected to be installed OpenJDK.

dgolovin:
test this

dgolovin:
test this

dgolovin:
test this

PR #1275, opened 2018-02-22
Use machine path rather than user path on windows

jrichter1:
Since we default to installing into program files, it seemed logical to add items to the system (machine) path rather than the user one. Things like oc still remain on user path since it is installed in user's home.

### DIFF
--- a/browser/model/cdk.js
+++ b/browser/model/cdk.js
@@ -70,7 +70,9 @@ class CDKInstall extends InstallableItem {
     }).then(()=> {
       return Platform.makeFileExecutable(ocExe);
     }).then(()=> {
-      return Platform.addToUserPath([ocExe, this.minishiftExe]);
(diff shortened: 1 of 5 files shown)


dgolovin:
Should work, except a scenario where target folder is changed to something that only visible for current user like ~/DevelopmentSuite.

I don't really see the value in this change. We basically let one user to pollute system environment with something we don't event know other users wanted.

jrichter1:
It's no big deal either way.
The two reasons I thought this might be of some use were:
 - jdk is added to machine path automatically, so why not other components
 - machine path takes precedence, so if there is an older version component already sitting on machine path, it effectively still acts as a default, even if we install a new one and put it on user path

In the end, it's your call :)  If you think this is undesired, just close the PR.

dgolovin:
Those are good points @jrichter1.

PR #1353, opened 2018-04-03
Change packaging for macOS to DMG

### LINKED ISSUE #1158: MacOS install should use .dmg
When doing the full download of the installer for MacOS with RH DevSuite, the download itself is 1.2GB zipped. This then has to be unzipped, which takes up another 1.3GB on disk, for 2.5GB of disk space usage before anything is even installed.

If the installer were a DMG, it not only would not require the duplication of disk space usage, it would permit a helpful tour to install and get started. Compare to the VirtualBox installer:

image

Links could go to the documentation, for example.

dgolovin:
Fix #1158 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ the GNU LGPL license, more details on <https://www.7zip.org>).
 
 #### macOS
 
-Both Installers are zipped macOS Application Bundles.
+Each Installer is a disk image (.dmg file) with macOS Application Bundle inside.
 
 Architecture
 ------------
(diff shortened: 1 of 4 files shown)


jrichter1:
Looks like no functionality gets broken by this (yay), so I'm trying to sum up what else would need to be changed (meh). So far I got this (anything missing?):
 - the build job (the publishing part)
 - docs
 - some of my scripts (which I can live with)
So if all the parties involved get updated, it should be good to go.

dgolovin:
@jrichter1 We should put our build script to git as well, so we can fix them at the same time.

dgolovin:
@jrichter1 this also opens opportunity to create shell script or app to uninstall on mac and put it next to installer inside dmg.

dgolovin:
@jrichter1 I've tested installation for bundled installer and it worked for me, should we merge this PR?

jrichter1:
system test

jrichter1:
system test

jrichter1:
system test

jrichter1:
system test

jrichter1:
system test

jrichter1:
system test

jrichter1:
looks like we are good to go

PR #1384, opened 2018-04-12
Allow Java 9+ to be detected

### LINKED ISSUE #1336: JDK not found on macOS
The installer doesn't see the JDK that's already installed on the machine. It's still displaying the message 'You have selected Red Hat JBoss Developer Studio, which requires Oracle Java version 1.8.0 or higher. Due to licence limitations we may not install Java for you. Please install JDK version 1.8.0 downloaded from this location and restart the Installer to continue'.
It is still displayed after re-installation of JDK.

dgolovin:
Fix #1336 and at least show message that Java 1.8.0 is required.

### DIFF
--- a/browser/model/jdk-install.js
+++ b/browser/model/jdk-install.js
@@ -2,7 +2,8 @@
 
 let fs = require('fs-extra');
 let path = require('path');
-var rimraf = require('rimraf');
+let rimraf = require('rimraf');
+let semver = require('semver');
 
 
 import InstallableItem from './installable-item';
@@ -35,7 +36,7 @@ class JdkInstall extends InstallableItem {
   }
 
   detectExistingInstall() {
-    let versionRegex = /version\s"(\d+\.\d+\.\d+).*"/;
+    let versionRegex = /version\s"(.+)".*/;
     let versionRegex1 = /(\d+\.\d+\.\d+).*/;
     let command = 'java -XshowSettings';
     this.addOption('install', versionRegex1.exec(this.version)[1], this.installerDataSvc.jdkDir(), true);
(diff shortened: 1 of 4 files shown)


browser/model/jdk-install.js
```diff
@@ -34,7 +35,7 @@ class JdkInstall extends InstallableItem {
   }
 
   detectExistingInstall() {
-    let versionRegex = /version\s"(\d+\.\d+\.\d+)_.*"/;
+    let versionRegex = /version\s"(\d+\.\d+\.\d+).*"/;
```
jrichter1: This is not going to match the new java version string. For java 9+ the output for `java -version` is
`java version "$majorVersion" yyyy-mm-dd`, `$majorVersion` being the 9+

Also, if you want to compare `$majorVersion` to say `1.8.0` it doesn't work straight away, but semver seems to have some API to adjust it.

dgolovin: For windows release output is:
```shell
c:\p\rh\dpi>java -version
openjdk version "9.0.4.1-redhat"
OpenJDK Runtime Environment (build 9.0.4.1-redhat+11)
OpenJDK 64-Bit Server VM (build 9.0.4.1-redhat+11, mixed mode)
```
and installer would show

image

@mohitsuman is checking macOS output right now.

dgolovin: On mac it is:

``` shell
openjdk version "9.0.4"
OpenJDK Runtime Environment (build 9.0.4+11)
OpenJDK 64-Bit Server VM (build 9.0.4+11, mixed mode)
```
Installer shows messages below:

image

dgolovin: @jrichter1 so my guess you are testing on RHEL, aren't you?

dgolovin:
What does not work is Java 10, becuase
``` shell
$java -version
openjdk version "10" 2018-03-20
OpenJDK Runtime Environment 18.3 (build 10+46)
OpenJDK 64-Bit Server VM 18.3 (build 10+46, mixed mode)
```

dgolovin:
we would probably simplify version detection then to just match numbers inside "(.*)" and then use `semver.coerce()` method before comparing 
``` javascript
> require("semver").coerce("1.8.0_161-1-redhat").raw
'1.8.0'
> require("semver").coerce("10").raw
'10.0.0'
> require("semver").coerce("9.0.4.1").raw
'9.0.4'
```

browser/model/jdk-install.js
mohitsuman: yeah for macOS 10.13.4 , it shows:

```
dhcp193-9:~ mohitsuman$ java -version
java version "9.0.4"
Java(TM) SE Runtime Environment (build 9.0.4+11)
Java HotSpot(TM) 64-Bit Server VM (build 9.0.4+11, mixed mode)

dgolovin:
Added support for Java 10 detection as @jrichter1 suggested:

image

PR #1386, opened 2018-04-12
Select Open JDK 8 to install if required and Java 9+ detected

dgolovin:
requires https://github.com/redhat-developer-tooling/developer-platform-install/pull/1384 and https://github.com/redhat-developer-tooling/developer-platform-install/pull/1385 to work.

### DIFF
--- a/browser/model/jdk-install.js
+++ b/browser/model/jdk-install.js
@@ -86,7 +86,7 @@ class JdkInstall extends InstallableItem {
   }
 
   getMsiSearchScriptData() {
-    return 'REG QUERY HKLM\\Software\\Microsoft\\Windows\\CurrentVersion\\Uninstall /f "OpenJDK" /s';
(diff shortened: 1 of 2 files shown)


jrichter1:
This actually lets you attempt to re-install openjdk msi. Because the msi puts java on the end of the path, the pre-existing jdk9+ will still take precedence when detected.
So when you install the msi, the next run of the installer lets you do it again, which does virtually nothing (only one installation may exist). In fact, it forces you to "install" jdk again if you select something that depends on it.

The most interesting fact is, when installing items dependent on openjdk along with re-installing openjdk into a different folder, all java calls magically get redirected to the original openjdk location.

dgolovin:
@jrichter1 that is all true, because devsuite installer just takes first element detected on path and goes with it. This is the less intrusive fix to do right now. 
IMO we should merge it.

PR #1395, opened 2018-04-17
Fix 'clear all' deselecting components while dependents are still selected

jrichter1:
Turns out one can select a component with its dependencies in one channel, then switch to a different channel and use 'clear all' to simply deselect the dependencies without the original item. 

Example: select Java Development in Guided Development channel -> go to Java Development channel -> click Clear All -> everything in Java Dev channel gets deselected even though Guided Development is still selected

This PR should check if the deselected component still has something depending on it before committing.

### DIFF
--- a/browser/pages/selection/controller.js
+++ b/browser/pages/selection/controller.js
@@ -52,6 +52,7 @@ class SelectionController {
   }
 
   toggleSelection(type) {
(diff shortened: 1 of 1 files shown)


jrichter1:
test this

jrichter1:
test this

PR #1403, opened 2018-04-19
Look for xhyve in /usr/local/bin

jrichter1:
when run from dmg, the installer has its PATH limited to `/usr/bin:/bin:/usr/sbin:/sbin`
no idea why, but since xhyve is usually in `/usr/local/bin` we should look there as well


dgolovin:
I have also tested without  this fix and xhyve was not detected.

